PR #14, opened 2023-03-14
RSDK-2014 - Appease linter & go-test

kkufieta:
Ticket: https://viam.atlassian.net/browse/RSDK-2014

**What this PR does**
* Made changes to appease the linter & go-test

**Pre-work done (into the kk/rsdk-2014-implement-cartographer-model dev branch)**

* Copied/pasted [builtin.go](https://github.com/viamrobotics/rdk/blob/main/services/slam/builtin/builtin.go), [builtin_test.go](https://github.com/viamrobotics/rdk/blob/main/services/slam/builtin/builtin_test.go), and [cartographer_int_test.go](https://github.com/viamrobotics/rdk/blob/main/services/slam/builtin/cartographer_int_test.go) 
* View the main branch/PR here: https://github.com/viamrobotics/viam-cartographer/pull/15


viam-cartographer.go
```diff
@@ -930,7 +946,7 @@ func (slamSvc *builtIn) getAndSaveDataDense(ctx context.Context, cams []camera.C
 
 	var fileType string
 	switch slamSvc.slamMode {
-	case slam.Dim2d:
+	case slam.Dim2d, slam.Dim3d:
```
zaporter-work, line 949: Should we throw an error here with "Dim3d is not implemented" like we do in CreateTimestampFile?

zaporter-work (approved):

viam-cartographer.go
kkufieta: Yes. Thank you for pointing this out.

.gitignore
```diff
@@ -17,3 +17,7 @@ build/
 
 # Code editor settings
 .vscode
+   
+# artifact
```
EshaMaharishi, line 21: [note to self] This is the same as in [rdk's .gitignore](https://github.com/viamrobotics/rdk/blob/fcae3e91d937d54cf84bd1ee80374b09898cb7f5/.gitignore#L60-L62)

.artifact/config.json
```diff
@@ -0,0 +1,12 @@
+{
```
EshaMaharishi, line 1: [note to self] same as in rdk

.artifact/tree.json
```diff
@@ -0,0 +1,910 @@
+{
+  "rimage": {
```
EshaMaharishi, line 2: Do you mind explaining the differences between this and [rdk's .artifact/tree.json](https://github.com/viamrobotics/rdk/blob/fcae3e91d937d54cf84bd1ee80374b09898cb7f5/.artifact/tree.json)?

viam-cartographer.go
```diff
@@ -281,6 +281,7 @@ func (slamSvc *builtIn) Position(ctx context.Context, name string, extra map[str
 		pInFrame = referenceframe.NewPoseInFrame(resp.GetComponentReference(), spatialmath.NewPoseFromProtobuf(resp.GetPose()))
 		returnedExt = resp.Extra.AsMap()
 	} else {
+		//nolint:staticcheck
```
EshaMaharishi, line 284: To confirm, we added these for the same reason we [currently exclude](https://github.com/viamrobotics/rdk/blob/6ea6876000cd6910e9f16ca994c8b9bd70f7e215/etc/.golangci.yaml#L100-L104) staticcheck from rdk's services/slam/ directory?

Will we automatically remove these lines at some point in the future, or should we make a note to remove these lines in RSDK-1053 or create a separate ticket for it?

viam-cartographer.go
```diff
@@ -470,7 +477,9 @@ func (slamSvc *builtIn) GetInternalStateStream(ctx context.Context, name string)
 }
 
 // New returns a new slam service for the given robot.
-func New(ctx context.Context, deps registry.Dependencies, config config.Service, logger golog.Logger, bufferSLAMProcessLogs bool) (slam.Service, error) {
+func New(ctx context.Context, deps registry.Dependencies, config config.Service,
+	logger golog.Logger, bufferSLAMProcessLogs bool,
```
EshaMaharishi, line 481: To confirm, this change is to obey the line width limit?

Curious why there needs to be a comma after `bufferSLAMProcessLogs bool`?

.artifact/tree.json
kkufieta: I copied/pasted RDK's tree.json and deleted what I didn't need.

viam-cartographer.go
kkufieta: Great question. We need to manually remove these lines in the future. I created this ticket since I couldn't find an existing one: https://viam.atlassian.net/browse/RSDK-2339

viam-cartographer.go
kkufieta: That's correct. I looked up why a comma is needed: without the comma, the lexer will insert a semicolon after `bool` which is invalid: https://go.dev/doc/effective_go#semicolons

viam-cartographer.go
EshaMaharishi: Perfect, thanks!

viam-cartographer.go
EshaMaharishi: Got it. It looked a bit unusual because there two arguments on that line, and I think typically a trailing comma is only used if each argument is on its own line ([example](https://github.com/viamrobotics/viam-cartographer/blob/fe8feabb09e429c6e0d89378aae1557740dabcb4/viam-cartographer.go#L619)).

I couldn't find Go documentation saying that, but for python: https://peps.python.org/pep-0008/#when-to-use-trailing-commas.

Would you mind pinging one of the Golang experts (Tess showed me [this wiki](https://viam.atlassian.net/wiki/spaces/ENG/pages/9732097/Language+Framework+Reviewers) yesterday) to confirm this is okay, or if it's something we'd want to change in the linter?

If you can't get an answer quickly, feel free to push this PR; I made a note for myself to check in about it.

kkufieta: Sure! @edaniels what are your thoughts on the above?

EshaMaharishi: Cool, that makes sense to me. It'd be nice if the linter did/suggested this automatically - is there a place to file a ticket for that?

EshaMaharishi: Oh, ok haha. Sounds good!

edaniels: you'd need to file that against golangci-lint probably and whatever linter complained. it's probably not worth it imo

EshaMaharishi: I see, got it thanks!

tessavitabile (approved):

PR #20, opened 2023-03-15
RSDK-2027 - Set up cartographer integration tests automation in CI

kkufieta:
Ticket: https://viam.atlassian.net/browse/RSDK-2027

Done:
* Set up cartographer integration tests automation in CI
    * I am keeping running the same integration tests against the code in RDK until we are ready to delete builtin from RDK.

PR for testing the CI: TODO


tessavitabile (approved):

.github/workflows/test.yml
```diff
@@ -96,3 +96,8 @@ jobs:
       if: matrix.platform == 'linux/amd64'
       run: |
         sudo -u testbot bash -lc 'cd rdk/services/slam/builtin && sudo go test -v -run TestCartographerIntegration'
+
+    - name: Run viam-cartographer cartographer integration tests
```
EshaMaharishi, line 100: Just for my knowledge, in case you have context on these:

1. Why in rdk do we use a `test-integration` Makefile target that calls `go test -v -run TestCartographerIntegration`, then call that target from a github workflow, but here we directly call `go test` from the github workflow?

2. I am guessing here we don't need to `export APPIMAGE_EXTRACT_AND_RUN=1` before calling `go test` because we aren't creating appimages in viam-cartographer yet?

EshaMaharishi (approved):

zaporter-work (approved):

.github/workflows/test.yml
kkufieta: Great questions. I wondered about the same thing. I also assumed the difference was due to the appimage used when we call `make test-integration`. @JohnN193 likely has more context than I do.

PR #25, opened 2023-03-20
RSDK-2204 Resolution-Based Sampling of PointCloud

jeremyrhyde:
This PR changes from a max size-based sampling of the final pointcloud to a resolution-based sampling. This is done by controlling the resolution used by the submap painter as in the current implementation the final point cloud is produced by iterating over the pixels in the produced image. By varying the resolution, we can increase or decrease the number of pixels in the image and thus control the output resolution of our pointcloud.

Future work may be done to allow users to change this parameter either via the config, or via additional parameters in the GRPC request. An important thing to note is that, unlike the other parameters which need to be instantiated in cartographer upon startup, this variable is available to be varying DURING operation. This allows for possible frontend control of the resolution should it be desired.

JIRA Ticket: https://viam.atlassian.net/browse/RSDK-2204

**Testing:**
Current testing is based on size of file, the number of points, and image quality (although it is a bit hard to see). I also ran a brute force test on the smaller pcd files to go through all points and confirm the minimum distance. These values proved the resolution was working as anticipated.

Resolution: 0.01 (current value being used + size sampling)
Number of Points in Pointcloud: 19,208,510
Pointcloud size: 307.3MB
Minimum Distance: File too big for brute force test
Image: N/A (too big for Micheal Parks' PCD viewer)

Resolution: 0.05
Number of Points in Pointcloud: 768,390
Pointcloud size:  12.3MB
Minimum Distance: File too big for brute force test
Image: 
mapPC_r005
close up of point distribution 
mapPC_r005_closeup

Resolution: 0.1
Number of Points in Pointcloud: 198,176
Pointcloud size: 3.2MB
Minimum Distance: 0.09999999999999272 (compared to 0.1)
Image: 
mapPC_r01
close up of point distribution 
mapPC_r01_closeup

Resolution: 0.5
Number of Points in Pointcloud: 9,058
Pointcloud size: 145KB
Minimum Distance: 0.4999999999999927 (compared to 0.5)
Image: 
image
close up of point distribution unavailable as it looks like a black screen

Resolution: 1.0 
Number of Points in Pointcloud: 2,370
Pointcloud size: 38KB
Minimum Distance: 0.9999999999997 (compared to 1.0)
Image: N/A (too few points that are too far apart to be visible in Micheal Parks PCD/viewer)


viam-cartographer/src/slam_service/slam_service.h
```diff
@@ -209,8 +219,10 @@ class SLAMServiceImpl final : public SLAMService::Service {
     double rotation_weight = 1.0;
 
    private:
-    // kPixelSize defines the pixel size for drawing the jpeg map
-    const double kPixelSize = 0.01;
+    // resolution defines the area in meters that each pixel represent. This 
+    // is used by drawing the jpeg map as well as the resolution of the outputted 
+    // PCD
+    const double resolution = 0.05;
```
jeremyrhyde, line 225: [q] do we want to move the location of this variable?

viam-cartographer/src/slam_service/slam_service.h
```diff
@@ -53,7 +53,7 @@ static const int jpegQuality = 50;
 // Byte limit on GRPC, used to help determine sampling skip_count
 static const int maximumGRPCByteLimit = 32 * 1024 * 1024;
 // Byte limit for chunks on GRPC, used for streaming apis
-static const int maximumGRPCByteChunkSize = 64 * 1024;
+static const int maximumGRPCByteChunkSize = 1 * 1024 * 1024;
```
JohnN193, line 56: [q] does this need to be accounted for in builtin.go?

viam-cartographer/src/slam_service/slam_service.h
JohnN193: I'd say we should move it to the top with the other semi-configurable parameters

JohnN193 (approved):
LGTM, only thought would be tweaking the colors of the PCD for the PR to make your tests a bit clearer, but I don't think that is super necessary

JohnN193:
> Resolution: 1.0 (current standard) Number of Points in Pointcloud: 2,370 Pointcloud size: 38KB Minimum Distance: 0.9999999999997 (compared to 1.0) Image: N/A (too few points that are too far apart to be visible in Micheal Parks PCD/viewer)

[q] what does `current standard` mean here?

viam-cartographer/src/slam_service/slam_service.h
```diff
@@ -53,7 +53,7 @@ static const int jpegQuality = 50;
...
 // Coeffient to adjust the skip count for the PCD to ensure the file is within
 // grpc limitations. Increase the value if you expect dense feature-rich maps
 static const int samplingFactor = 1;
```
JohnN193, line 59: [minor] remove unused variables

viam-cartographer/src/slam_service/slam_service.h
```diff
@@ -65,6 +65,21 @@ static const Eigen::Quaterniond pcdRotation(0.7071068, -0.7071068, 0, 0);
 // This will result in rotations occurring within the y axis to match 2D mapping
 // in the XZ plane
 static const Eigen::Quaterniond pcdOffsetRotation(0.7071068, 0.7071068, 0, 0);
+
+// resolution defines the area in meters that each pixel represent. This
+// is used to draw the jpeg map and in so doing defines the resolution of
+// the outputted PCD
+const double resolution = 0.05;
```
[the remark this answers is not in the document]
jeremyrhyde: I have it set to 0.05 (5 cm) which is different from the original resolution of 0.01 (1 cm) as this produced a 20,000,000 point pointcloud (300 MB) for a run of part of the office. With streaming, we could handle this but the latency is bad as it takes about 30 secs to paint and process. The 0.05 I have currently produces a 12.3MB file with 768,390 points. 

If desired I can do a time analysis and choose a resolution to hit a benchmark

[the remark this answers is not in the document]
tessavitabile: Can you let Ray know what the resolution is, and also let him know we can configure this?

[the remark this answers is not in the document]
jeremyrhyde: I spoke with him and he thinks a 5 cm resolution is a good baseline for motion currently but if he needs a different resolution in the interim, he can ping us and we can shoot off a change. He also has been made aware that we will be making this dynamic in the future potentially via the addition of a `GetPointCloudMap` request parameter.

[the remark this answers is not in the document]
tessavitabile: great, thank you!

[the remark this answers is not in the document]
JohnN193: would it be worth adding it as a `config_param` in this PR? it would require no changes to rdk/builtin

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -579,77 +579,55 @@ void SLAMServiceImpl::GetLatestSampledPointCloudMapString(
...
-        viam::utils::writeFloatToBufferInBytes(data_buffer,
-                                               rotated_map_point.z());
-        viam::utils::writeIntToBufferInBytes(data_buffer, prob);
+    for (int pixel_y = 0; pixel_y < height; pixel_y++) {
+        for (int pixel_x = 0; pixel_x < width; pixel_x++) {
+            // Get bytes associated with pixel
+            int pixel_index = pixel_x + pixel_y * width;
+            int byte_index = pixel_index * bytesPerPixel;
+            std::vector<unsigned char> pixel_data = {
+                data + byte_index, data + byte_index + bytesPerPixel};
```
EshaMaharishi, line 602: [note to self] pointer arithmetic

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -660,11 +638,20 @@ void SLAMServiceImpl::GetLatestSampledPointCloudMapString(
...
+           (pixel_data[1] == defaultCairosEmptyPaintedSlice) &&
+           (pixel_data[2] == defaultCairosEmptyPaintedSlice);
+}
+
+int calculateViamProbabilityFromColorChannel(
+    std::vector<unsigned char> pixel_data) {
     unsigned char maxVal = CHAR_MAX;
     unsigned char minVal = defaultCairosEmptyPaintedSlice;
     unsigned char maxProb = 100;
     unsigned char minProb = 0;
```
EshaMaharishi, line 652: [note to self] this is where we convert to the 0-100 range

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -579,77 +579,55 @@ void SLAMServiceImpl::GetLatestSampledPointCloudMapString(
...
+
+            // Skip pixel if it contains empty data (default color)
+            if (checkIfEmptyPixel(pixel_data)) continue;
+
+            // Determine probability based on color pixel and skip if 0
+            int prob = calculateViamProbabilityFromColorChannel(pixel_data);
+            if (prob == 0) continue;
+
+            // Convert pixel location to pointcloud point in meters
+            float x_pos = (pixel_x - painted_slices->origin.x()) * resolution;
```
EshaMaharishi, line 612: [note to self] resolution is in meters. so resolution = 3 means you want a pixel every 3 meters.

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -740,7 +727,7 @@ SLAMServiceImpl::GetLatestPaintedMapSlices() {
             &submap_slice.cairo_data);
     }
     cartographer::io::PaintSubmapSlicesResult painted_slices =
-        viam::io::PaintSubmapSlices(submap_slices, kPixelSize);
+        viam::io::PaintSubmapSlices(submap_slices, resolution);
```
EshaMaharishi, line 730: [q] if we have an image of 8 pixels with values like

[20, 20, 20, 20, 20, 0, 0, 0, 0]

and want to downsample it to 2 pixels, will it still show some non-zero value for the right-hand pixel?

e.g., either

[20, 5] // use average of underlying pixels

or

[20, 20] // use max of underlying pixels

EshaMaharishi (approved):
LGTM! I just had one question for my own knowledge

viam-cartographer/src/slam_service/slam_service.h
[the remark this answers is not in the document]
kkufieta: [nit] Change `resolution` to `resolution_meters`

viam-cartographer/src/slam_service/slam_service.h
```diff
@@ -47,24 +47,34 @@ namespace viam {
...
 // Quaternion to rotate axes to the XZ plane
 static const Eigen::Quaterniond pcdRotation(0.7071068, -0.7071068, 0, 0);
 // Static offset quaternion, so orientation matches physical intuition.
 // This will result in rotations occurring within the y axis to match 2D mapping
 // in the XZ plane
 static const Eigen::Quaterniond pcdOffsetRotation(0.7071068, 0.7071068, 0, 0);
+
+// resolution defines the area in meters that each pixel represent. This
+// is used to draw the jpeg map and in so doing defines the resolution of
+// the outputted PCD
```
kkufieta, line 65: [nit] Uppercase to match the style of the rest of the document
```suggestion
// Resolution defines the area in meters that each pixel represents. This
// is used to draw the jpeg map and, in so doing, defines the resolution of
// the outputted PCD
```

viam-cartographer/src/slam_service/slam_service.h
```diff
@@ -47,24 +47,34 @@ namespace viam {
...
+const double resolution = 0.05;
+// Number of bytes in a pixel
+const int bytesPerPixel = 4;
+// Color channel (RGBA) used to determine probability of point
+// - 0 is the R channel
+// - 1 is the B channel
+// - 2 is the G channel
+// - 3 is the A channel
+const int probabilityColorChannel = 2;
+// This RGB value represents "no input" from a Cario pained map
```
kkufieta, line 75: [q] What does "no input" mean? What does the value of `102` represent? One of the RGB values?

viam-cartographer/src/slam_service/slam_service.h
```diff
@@ -74,10 +84,15 @@ using SensorId = cartographer::mapping::TrajectoryBuilderInterface::SensorId;
...
-// from cartographer
-unsigned char ViamColorToProbability(unsigned char color);
+// For a given color channel (probabilityColorChannel) convert the scale from
+// the given 102-255 range to 100-0. This is an initial solution for extracting
+// probability information from cartographer
+int calculateViamProbabilityFromColorChannel(
+    std::vector<unsigned char> pixel_data);
+
+// Check if pixel is not in our map and is the default color (i.e. RGB =
+// [102,102,102])
```
kkufieta, line 94: [q] What does "Check if pixel is not in our map" mean? What does "default color" mean?

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -579,77 +579,55 @@ void SLAMServiceImpl::GetLatestSampledPointCloudMapString(
         }
     }
 
+    // Get data from painted surface in RGBA format
     auto painted_surface = painted_slices->surface.get();
     int width = cairo_image_surface_get_width(painted_surface);
     int height = cairo_image_surface_get_height(painted_surface);
-    // Get all pixels from the painted surface in RGBA format
     auto data = cairo_image_surface_get_data(painted_surface);
```
kkufieta, line 586: [nit] Please use the type instead of `auto` for improved readability. I recommend [opt] to rename `data` to `surface_data` as well.

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -579,77 +579,55 @@ void SLAMServiceImpl::GetLatestSampledPointCloudMapString(
...
+
+            // Convert pixel location to pointcloud point in meters
+            float x_pos = (pixel_x - painted_slices->origin.x()) * resolution;
+            // Y is inverted to match output from getPosition()
+            float y_pos = -(pixel_y - painted_slices->origin.y()) * resolution;
+            float z_pos = 0;  // Z is 0 in 2D SLAM
+
+            // Add point to buffer
+            Eigen::Vector3d map_point(x_pos, y_pos, z_pos);
+            auto rotated_map_point = pcdRotation * map_point;
```
kkufieta, line 619: [nit] Please use the type instead of `auto`.

kkufieta (changes requested):
Nice work! I have a few [nit]s and questions.

Regarding using `auto`: It should be used sparingly and only if it is easy to understand what type it stands for. See this article as a reference: https://stackoverflow.com/questions/6900459/the-new-keyword-auto-when-should-it-be-used-to-declare-a-variable-type/6900522#6900522

tessavitabile:
At MongoDB, we used `auto` wherever possible. I'm not 100% certain I remember the justification, but I think it was to avoid having to change a lot of code if you changed a type name, and maybe to avoid implicit type conversion? It might be interesting to ask in #project-sdks-bare-metal.

viam-cartographer/src/slam_service/slam_service.h
jeremyrhyde: I want to keep resolution lowercase to make it clear we are talking about the variable, however, I will add a word in from to be capitalized

kkufieta:
@tessavitabile I don't think changing a type name will be a concern in the case of cartographer. Also, when looking at one of the places I commented on using the type instead of `auto`, you can see that Esha also had to add a comment for herself to understand the code. I had to look up the function definition, see what it returns, and then understand the code. The Stackoverflow article represents this experience of how `auto` reduced the readability well, which is why I suggest that we should follow that guideline.

An alternative when we'd instead use `auto` is making names more explicit to reflect the type. This would help avoid having to dig into the code to understand it.

One more argument is that the style of cartographer has been mainly to avoid using `auto`. Diverting from that would break the consistency of the code. I don't mind changing the approach in other projects, but I think consistency should also be a factor here.

viam-cartographer/src/slam_service/slam_service.h
kkufieta: I see - that wasn't clear to me. Yeah, maybe `The resolution variable defines...` would have been less confusing for me.

viam-cartographer/src/slam_service/slam_service.h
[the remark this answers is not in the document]
jeremyrhyde: Interesting idea John! Eventually, it most likely will be added as a field in the `GetPointCloudMapRequest` message but this could allow us to manipulate it in the interim. Benefits would include being able to control the size of pointclouds being sent over the wire until PCD deltas are available or API changes are made. 

@EshaMaharishi @tessavitabile what do you think?

viam-cartographer/src/slam_service/slam_service.cc
jeremyrhyde: So this is not downsampling the image produced by cairo. Cairo combines the submaps and more accurately their associated probability grids into an image. We then send all the pixels in that image over that arent empty or have a prob of 0. 

So therefore in the image produced by cario has pixel values of [20, 20, 20, 20, 20, 0, 0, 0] we would send over [f(20), f(20), f(20), f(20), f(20)] where the function f() is `calculateViamProbabilityFromColorChannel`.

If instead, you are asking how cairo averages the probability in overlapping submaps/probability grids, the answer is that we are not sure exactly how they are doing this and if they are taking the maximum, average, or some other weighted average

viam-cartographer/src/slam_service/slam_service.h
jeremyrhyde: no input refers to pixels that the painter has not adjusted based on the submaps (i.e. the default background).

102 is the default RGB value each pixel will have. rather than starting at 0. This 102 is also used in the probability calculation. A pixel that has all 3 channels as the default value([102,102,102]) will result in the grey pixels outside of the map.

Let me know if the adjustment to the comment makes that clear or if you'd like something more

viam-cartographer/src/slam_service/slam_service.h
jeremyrhyde: "pixel not in our map" refers to the grey pixels that represent the background color. We want to filter out these pixels. Default color is `defaultCairosEmptyPaintedSlice`, more explicitly when all 3 color channels have the default value(resulting in the grey pixels) 

I changed the comment to better reflect this. Take a look!

viam-cartographer/src/slam_service/slam_service.cc
jeremyrhyde: I agree about wanting to be type-rigid in our code but not only does cartographer uses `auto` heavily, particularly when it comes to their custom types and `cairo` types but that is our current pattern as well. I personally think it's better to stick to both current patterns and bring up the topic of refactoring for type rigidity during full modularization.

I do agree though that the `rotated_map_point` should be defined since `Eigen:Vector3d` is a much more ubiquitous type.

tessavitabile:
I don't have a strong opinion here, just wanted to point out that I don't think there's consensus in the C++ development world on when to use auto. Thanks for posting in #project-sdks-bare-metal. It looks like we didn't get input from anyone outside the team. I'm fine with whatever the team prefers.

viam-cartographer/src/slam_service/slam_service.h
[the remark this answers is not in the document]
EshaMaharishi: I think it's a good idea to make it configurable in the future, but it's ok to push this PR with it hard-coded as a first step

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -579,77 +579,58 @@ void SLAMServiceImpl::GetLatestSampledPointCloudMapString(
         }
     }
 
+    // Get data from painted surface in RGBA format
```
jeremyrhyde, line 582: @EshaMaharishi This is where we use the painted submaps (a cairo object) and convert it to usable pixel data for serialization.

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -740,7 +730,7 @@ SLAMServiceImpl::GetLatestPaintedMapSlices() {
             &submap_slice.cairo_data);
     }
     cartographer::io::PaintSubmapSlicesResult painted_slices =
-        viam::io::PaintSubmapSlices(submap_slices, kPixelSize);
+        viam::io::PaintSubmapSlices(submap_slices, resolution_meters);
```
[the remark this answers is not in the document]
jeremyrhyde: painted submap:

```
struct PaintSubmapSlicesResult {
  PaintSubmapSlicesResult(::cartographer::io::UniqueCairoSurfacePtr surface,
                          Eigen::Array2f origin)
      : surface(std::move(surface)), origin(origin) {}
  ::cartographer::io::UniqueCairoSurfacePtr surface;

  // Top left pixel of 'surface' in map frame.
  Eigen::Array2f origin;
};
```

viam-cartographer/src/slam_service/slam_service.cc
jeremyrhyde: As discussed offline, the "cairo combined map -(2)-> cairo combined map using fewer pixels" is a combined step as we send cairo the submap data AND resolution at the same time. The generated painted submaps are the result of that call.

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -579,77 +579,58 @@ void SLAMServiceImpl::GetLatestSampledPointCloudMapString(
...
-                                               rotated_map_point.y());
-        viam::utils::writeFloatToBufferInBytes(data_buffer,
-                                               rotated_map_point.z());
-        viam::utils::writeIntToBufferInBytes(data_buffer, prob);
+    for (int pixel_y = 0; pixel_y < height; pixel_y++) {
+        for (int pixel_x = 0; pixel_x < width; pixel_x++) {
+            // Get bytes associated with pixel
+            int pixel_index = pixel_x + pixel_y * width;
+            int byte_index = pixel_index * bytesPerPixel;
+            std::vector<unsigned char> pixel_data = {
```
nicksanford, line 601: I might be reading this wrong but it looks like `pixel_data` is always a vector with 2 elements, but are accessing 3 elements in the vector: https://github.com/viamrobotics/viam-cartographer/pull/25/files#diff-92ff8a5cd3a57430d78c6ba3b61a91d1810399b77613eece3c55841868d2822eR644. I don't understand how this can work. What am I missing?

nicksanford (changes requested):

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -660,11 +641,20 @@ void SLAMServiceImpl::GetLatestSampledPointCloudMapString(
...
+}
+
+int calculateViamProbabilityFromColorChannel(
+    std::vector<unsigned char> pixel_data) {
     unsigned char maxVal = CHAR_MAX;
     unsigned char minVal = defaultCairosEmptyPaintedSlice;
     unsigned char maxProb = 100;
     unsigned char minProb = 0;
+
+    unsigned char color = (int)pixel_data[probabilityColorChannel];
```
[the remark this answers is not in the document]
nicksanford: Why do we need to feed in the entire `pixel_data` if we are only interested in the `color` part of it?

viam-cartographer/src/slam_service/slam_service.cc
nicksanford: Also, I don't understand why this needs to be a vector at all. It appears as if we are only interested in the probability information, which should be a single `unsigned char` right?

EshaMaharishi: Great catch! This is using curly bracket initialization, which means it's creating a vector of 2 elements. The program isn't terminating when accessing the 3rd element because std::vector's operator[] is not bound-checked: https://cplusplus.com/reference/vector/vector/operator[]/.

I think we might have meant to use parentheses, like we did before:

```
std::vector<unsigned char> data_vect(data, data + size_data);
```

and as described in example 4 here: https://www.geeksforgeeks.org/initialize-a-vector-in-cpp-different-ways/.

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -579,77 +579,58 @@ void SLAMServiceImpl::GetLatestSampledPointCloudMapString(
...
+
+            // Determine probability based on color pixel and skip if 0
+            int prob = calculateViamProbabilityFromColorChannel(pixel_data);
+            if (prob == 0) continue;
+
+            // Convert pixel location to pointcloud point in meters
+            float x_pos =
+                (pixel_x - painted_slices->origin.x()) * resolution_meters;
+            // Y is inverted to match output from getPosition()
+            float y_pos =
```
nicksanford, line 616: Why is Y inverted in GetPosition?

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -579,77 +579,58 @@ void SLAMServiceImpl::GetLatestSampledPointCloudMapString(
...
-    // data_vect[i + 2] is the G channel
-    // data_vect[i + 3] is the A channel
-    std::vector<unsigned char> data_vect(data, data + size_data);
-
+    auto image_data = cairo_image_surface_get_data(painted_surface);
+    // // Each pixel contains 4 bytes of information in RGBA format
+    // // data_vect[i + 0] is the R channel
+    // // data_vect[i + 1] is the B channel
+    // // data_vect[i + 2] is the G channel
+    // // data_vect[i + 3] is the A channel
```
EshaMaharishi, line 591: Since `data_vect` no longer exist, should this comment move somewhere else?

Also, we can probably remove the double `//`

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -579,77 +579,58 @@ void SLAMServiceImpl::GetLatestSampledPointCloudMapString(
...
+            // Skip pixel if it contains empty data (default color)
+            if (checkIfEmptyPixel(pixel_data)) continue;
+
+            // Determine probability based on color pixel and skip if 0
+            int prob = calculateViamProbabilityFromColorChannel(pixel_data);
+            if (prob == 0) continue;
+
+            // Convert pixel location to pointcloud point in meters
+            float x_pos =
+                (pixel_x - painted_slices->origin.x()) * resolution_meters;
```
nicksanford, line 614: I don't understand this formula, could you please explain it?

nicksanford: 1. Is `painted_slices->origin.x()` always the same in every iteration? 
2. Is calling `painted_slices->origin.x()` expensive or is it just returning a struct or object member? 

I ask b/c if it is both always the same but expensive to compute we might want to assign it to a variable outside of the loop & use that variable here rather than de-referencing the pointer & calling the method for every pixel in the image.

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -579,77 +579,58 @@ void SLAMServiceImpl::GetLatestSampledPointCloudMapString(
...
+            std::vector<unsigned char> pixel_data = {
+                image_data + byte_index,
+                image_data + byte_index + bytesPerPixel};
+
+            // Skip pixel if it contains empty data (default color)
+            if (checkIfEmptyPixel(pixel_data)) continue;
+
+            // Determine probability based on color pixel and skip if 0
+            int prob = calculateViamProbabilityFromColorChannel(pixel_data);
+            if (prob == 0) continue;
```
nicksanford, line 610: I would like to propose we not use if statements w/o brackets as it increases the risk that someone will add a new clause to the if statement which they believe will execute if the predicate is true but actually happen at all times.

This is a bug from apple which compromised their TLS security which was caused by the foot gun than bracketless if statements can cause: https://www.imperialviolet.org/2014/02/22/applebug.html

@EshaMaharishi Curious do you have any opinion on the matter?

nicksanford (changes requested):

viam-cartographer/src/slam_service/slam_service.cc
JohnN193: its more the painter ends up painting Y inverted, so we flip that here to change back to the orientation cartographer's internal state/position uses.

viam-cartographer/src/slam_service/slam_service.cc
EshaMaharishi: Agree, I would prefer always using brackets

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -579,77 +579,58 @@ void SLAMServiceImpl::GetLatestSampledPointCloudMapString(
...
-    // MB, with additional tuning provided by the samplingFactor. If the PCD
-    //  would already be smaller than 32MB, do not sample the image.
-    // When moving to streaming this behavior may change.
-    int skip_count = (size_data * pixelBytetoPCDByte) / maximumGRPCByteLimit *
-                     samplingFactor;
-    if (skip_count == 0) {
-        skip_count = 1;
-    }
-
     std::string data_buffer;
```
EshaMaharishi, line 595: Could we rename this to `pcd_data` to make it slightly more clear what it contains?

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -579,77 +579,58 @@ void SLAMServiceImpl::GetLatestSampledPointCloudMapString(
...
-    int size_data = width * height * 4;
-
-    // Each pixel contains 4 bytes of information in RGBA format
-    // data_vect[i + 0] is the R channel
-    // data_vect[i + 1] is the B channel
-    // data_vect[i + 2] is the G channel
-    // data_vect[i + 3] is the A channel
-    std::vector<unsigned char> data_vect(data, data + size_data);
-
+    auto image_data = cairo_image_surface_get_data(painted_surface);
```
EshaMaharishi, line 586: Could we check the returned format is what we expect via https://www.cairographics.org/manual/cairo-Image-Surfaces.html#cairo-image-surface-get-format?

viam-cartographer/src/slam_service/slam_service.cc
EshaMaharishi, line 591: Could we confirm this format is what we expect? We think it should be one of these: https://www.cairographics.org/manual/cairo-Image-Surfaces.html#cairo-format-t

viam-cartographer/src/slam_service/slam_service.cc
jeremyrhyde: I'm happy to change this to a more explicit format however, the code as it stands is doing as intended. When I print out the size and content of pixel_data I get:

`Size of pixel_data: 4 | [102 102 102 204 ]`

Now, this is a bit confusing because this appears to be in the RGBA format not ARGB. But that is a separate issue which I'm still investigating

viam-cartographer/src/slam_service/slam_service.h
```diff
@@ -47,24 +47,32 @@ namespace viam {
...
 static const Eigen::Quaterniond pcdRotation(0.7071068, -0.7071068, 0, 0);
 // Static offset quaternion, so orientation matches physical intuition.
 // This will result in rotations occurring within the y axis to match 2D mapping
 // in the XZ plane
 static const Eigen::Quaterniond pcdOffsetRotation(0.7071068, 0.7071068, 0, 0);
+
+// The resolution_meters variable defines the area in meters that each pixel
+// represents. This is used to draw the jpeg map and in so doing defines the
+// resolution of the outputted PCD
+const double resolution_meters = 0.05;
```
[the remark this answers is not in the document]
jeremyrhyde: Discussed offline. Delaying a decision on this until we expose this to the user. For now, also keeping it as meters because that is cartographer's scale/

viam-cartographer/src/slam_service/slam_service.cc
jeremyrhyde: Investigation complete. We are indeed using the ARGB format type, however, due to the big-endian encoding the bytes are arranged as BGRA, meaning we have been using the red channel this entire time for the probability info. The reason we never saw red in the images was due to the JPEG to a string converter setting the pixel in the image to RGB24, see [viam-cartographer/viam-cartographer/src/io/cairo_jpg.cc](https://github.com/viamrobotics/viam-cartographer/blob/main/viam-cartographer/src/io/cairo_jpg.cc)

jeremyrhyde: I have added a color struct to reduce confusion in the future.

viam-cartographer/src/slam_service/slam_service.cc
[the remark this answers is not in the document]
jeremyrhyde: Outdated

viam-cartographer/src/slam_service/slam_service.cc
jeremyrhyde: origin.x() should not change. However @JohnN193 can you confirm what coordinates/frame the origin frame are in? based on the comment in the carto code regarding this:

`// Get top left pixel of 'surface' in map frame`

and my own intuition on how this is being used. Should it in fact be 

`float x_pos = pixel_x * resolution_meters - painted_slices->origin.x()`

viam-cartographer/src/slam_service/slam_service.cc
jeremyrhyde: added!

viam-cartographer/src/slam_service/slam_service.cc
jeremyrhyde: Done!

viam-cartographer/src/slam_service/slam_service.cc
jeremyrhyde: Done!

viam-cartographer/src/slam_service/slam_service.cc
jeremyrhyde: Outdated

EshaMaharishi:
@jeremyrhyde I'm assuming you'll re-request review when you're ready?

viam-cartographer/src/slam_service/slam_service.cc
jeremyrhyde: Origin is indeed in pixel units as can be seen in its definition above and the use of `kPaddingPixel`. Therefore our equation is: 
```
float origin_map = (0-origin)*res // origin is in pixels
float x_pos = (pixel_x - 0) * res + origin_map // origin_map is in meters
float y_pos = -((pixel_y - 0) * res + origin_map) 
```
which can be simplified to:
```
float x_pos = (pixel_x - 0) * res +(0 - origin)*res = (pixel_x - origin)*res
float y_pos = -((pixel_y - 0) * res + (0 - origin)*res) = -(pixel_y - origin)*res
```

If you want we can convert the origin definition outside the loop to meters immediately

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -468,94 +468,97 @@ void SLAMServiceImpl::GetLatestSampledPointCloudMapString(
...
     int num_points = 0;
+    std::string pcd_data;
+    for (int pixel_y = 0; pixel_y < height; pixel_y++) {
+        for (int pixel_x = 0; pixel_x < width; pixel_x++) {
+            // Get byte index associated with pixel
+            int pixel_index = pixel_x + pixel_y * width;
+            int byte_index = pixel_index * bytesPerPixel;
+
+            // 4 bytes of ARGB color information are in reverse order due to
+            // little endian encoding
```
EshaMaharishi, line 498: Suggestion to say "We assume we are running on a little-endian system, so the ARGB order is reversed."

EshaMaharishi (approved):
LGTM mod comments

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -19,6 +19,44 @@
...
+    unsigned char G;
+    unsigned char B;
+};
+
+// Check if pixel is the default color signalling no data is present
+bool checkIfEmptyPixel(pixelColorARGB color) {
+    return (color.R == defaultCairosEmptyPaintedSlice) &&
+           (color.G == defaultCairosEmptyPaintedSlice) &&
+           (color.B == defaultCairosEmptyPaintedSlice);
+}
```
kkufieta, line 41: [nit]
```suggestion
bool CheckIfEmptyPixel(ColorARGB pixel_color) {
    return (pixel_color.R == defaultCairosEmptyPaintedSlice) &&
           (pixel_color.G == defaultCairosEmptyPaintedSlice) &&
           (pixel_color.B == defaultCairosEmptyPaintedSlice);
}
```

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -333,97 +371,81 @@ void SLAMServiceImpl::GetLatestSampledPointCloudMapString(
...
+            pixelColorARGB color;
+            color.A = image_data[byte_index + 3];
+            color.R = image_data[byte_index + 2];
+            color.G = image_data[byte_index + 1];
+            color.B = image_data[byte_index + 0];
+
+            // Skip pixel if it contains empty data (default color)
+            if (checkIfEmptyPixel(color)) {
+                continue;
+            }
```
kkufieta, line 411: [nit]
```suggestion
            ColorARGB pixel_color;
            pixel_color.A = image_data[byte_index + 3];
            pixel_color.R = image_data[byte_index + 2];
            pixel_color.G = image_data[byte_index + 1];
            pixel_color.B = image_data[byte_index + 0];

            // Skip pixel if it contains empty data (default color)
            if (CheckIfEmptyPixel(pixel_color)) {
                continue;
            }
```

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -333,97 +371,81 @@ void SLAMServiceImpl::GetLatestSampledPointCloudMapString(
...
-        // 2D SLAM so Z is set to 0
-        float z_pos = 0;
-
-        // Turn the map point into a vector to perform transformations with.
-        // Current transformation rotates coordinates to match slam service
-        // expectation (XZ plane)
-        Eigen::Vector3d map_point(x_pos, y_pos, z_pos);
-        auto rotated_map_point = pcdRotation * map_point;
+            // Determine probability based on color pixel and skip if 0
+            int prob = calculateProbabilityFromColorChannels(color);
```
kkufieta, line 414:
```suggestion
            // Determine probability based on the pixel color and skip if it is 0
            int prob = CalculateProbabilityFromColorChannels(pixel_color);
```

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -872,4 +894,4 @@ double SLAMServiceImpl::GetRotationWeightFromMapBuilder() {
     return map_builder.GetRotationWeight();
 }
 
-}  // namespace viam
+}  // namespace viam
```
kkufieta, line 897: [nit] Add newline

kkufieta (approved):
LGTM mod nit and question

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -468,94 +468,97 @@ void SLAMServiceImpl::GetLatestSampledPointCloudMapString(
...
+        throw std::runtime_error(errorLog);
+    }
     int width = cairo_image_surface_get_width(painted_surface);
     int height = cairo_image_surface_get_height(painted_surface);
-    // Get all pixels from the painted surface in RGBA format
-    auto data = cairo_image_surface_get_data(painted_surface);
-
-    // Total number of bytes in image (4 bytes per pixel)
-    int size_data = width * height * 4;
+    auto image_data = cairo_image_surface_get_data(painted_surface);
```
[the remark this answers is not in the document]
jeremyrhyde: I like this name change

viam-cartographer/src/slam_service/slam_service.h
```diff
@@ -43,24 +44,33 @@ namespace viam {
...
```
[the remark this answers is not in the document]
jeremyrhyde: Abiding my camel case for const values

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -333,97 +371,81 @@ void SLAMServiceImpl::GetLatestSampledPointCloudMapString(
...
     }
 
+    // Get data from painted surface in ARGB32 format
     auto painted_surface = painted_slices->surface.get();
+    auto image_format = cairo_image_surface_get_format(painted_surface);
+    if (image_format != cartographer::io::kCairoFormat) {
+        std::string errorLog =
+            "Error cairo surface in wrong format, expected Cairo_Format_ARGB32";
+        LOG(ERROR) << errorLog;
+        throw std::runtime_error(errorLog);
```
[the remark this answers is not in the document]
jeremyrhyde: Need to be changed in other places in separate PR

viam-cartographer/src/slam_service/slam_service.h
```diff
@@ -43,24 +44,33 @@ namespace viam {
...
+// Number of bytes in a pixel
+const int bytesPerPixel = 4;
```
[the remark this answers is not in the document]
jeremyrhyde: They can be! And will be shortly

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -19,6 +19,44 @@
 #include "cartographer/mapping/map_builder.h"
 #include "glog/logging.h"
 
+namespace {
+// The default value Cairo when coloring the image if no data is present. A
```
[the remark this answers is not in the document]
jeremyrhyde: removing cairo from comment as we actually set this default in the submap_painter code

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -19,6 +19,44 @@
...
+// Number of bytes in a pixel
+const int bytesPerPixel = 4;
+struct pixelColorARGB {
+    unsigned char A;
+    unsigned char R;
+    unsigned char G;
+    unsigned char B;
+};
+
+// Check if pixel is the default color signalling no data is present
```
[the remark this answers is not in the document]
jeremyrhyde: fixed!

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -19,6 +19,44 @@
...
+// For a given color channel (probabilityColorChannel) convert the scale from
+// the given 102-255 range to 100-0. This is an initial solution for extracting
+// probability information from cartographer
+int calculateProbabilityFromColorChannels(pixelColorARGB color) {
+    unsigned char maxVal = CHAR_MAX;
+    unsigned char minVal = defaultCairosEmptyPaintedSlice;
+    unsigned char maxProb = 100;
+    unsigned char minProb = 0;
+
+    // Probability is current determined solely by the R channel
```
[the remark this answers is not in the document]
jeremyrhyde: fixed!

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -19,6 +19,44 @@
...
+// Check if pixel is the default color signalling no data is present
+bool checkIfEmptyPixel(pixelColorARGB color) {
+    return (color.R == defaultCairosEmptyPaintedSlice) &&
+           (color.G == defaultCairosEmptyPaintedSlice) &&
+           (color.B == defaultCairosEmptyPaintedSlice);
+}
+
+// For a given color channel (probabilityColorChannel) convert the scale from
+// the given 102-255 range to 100-0. This is an initial solution for extracting
+// probability information from cartographer
```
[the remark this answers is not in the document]
jeremyrhyde: Removed reference

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -19,6 +19,44 @@
...
+    unsigned char maxVal = CHAR_MAX;
+    unsigned char minVal = defaultCairosEmptyPaintedSlice;
+    unsigned char maxProb = 100;
+    unsigned char minProb = 0;
+
+    // Probability is current determined solely by the R channel
+    unsigned char colorChannel = color.R;
+    unsigned char prob =
+        (maxVal - colorChannel) * (maxProb - minProb) / (maxVal - minVal);
+    return prob = std::min(std::max(prob, minProb), maxProb);
```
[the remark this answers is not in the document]
jeremyrhyde: return simplified

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -19,6 +19,44 @@
...
+}
+
+// For a given color channel (probabilityColorChannel) convert the scale from
+// the given 102-255 range to 100-0. This is an initial solution for extracting
+// probability information from cartographer
+int calculateProbabilityFromColorChannels(pixelColorARGB color) {
+    unsigned char maxVal = CHAR_MAX;
+    unsigned char minVal = defaultCairosEmptyPaintedSlice;
+    unsigned char maxProb = 100;
+    unsigned char minProb = 0;
```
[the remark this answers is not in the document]
jeremyrhyde: moving all variables non consts to snake case due to google C++ style guidelines

viam-cartographer/src/slam_service/slam_service.cc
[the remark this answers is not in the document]
jeremyrhyde: moving all variables non consts to snake case due to google C++ style guidelines

viam-cartographer/src/slam_service/slam_service.cc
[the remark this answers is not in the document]
jeremyrhyde: moving all variables non consts to snake case due to google C++ style guidelines

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -19,6 +19,44 @@
...
+bool checkIfEmptyPixel(pixelColorARGB color) {
+    return (color.R == defaultCairosEmptyPaintedSlice) &&
+           (color.G == defaultCairosEmptyPaintedSlice) &&
+           (color.B == defaultCairosEmptyPaintedSlice);
+}
+
+// For a given color channel (probabilityColorChannel) convert the scale from
+// the given 102-255 range to 100-0. This is an initial solution for extracting
+// probability information from cartographer
+int calculateProbabilityFromColorChannels(pixelColorARGB color) {
```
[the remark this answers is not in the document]
jeremyrhyde: done1

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -19,6 +19,44 @@
 #include "cartographer/mapping/map_builder.h"
 #include "glog/logging.h"
 
+namespace {
+// The default value Cairo when coloring the image if no data is present. A
+// pixel representing no data will have all 3 channels of its color channels
+// (RGB) as the default value([102,102,102])
+static const unsigned char defaultCairosEmptyPaintedSlice = 102;
```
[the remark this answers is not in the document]
jeremyrhyde: Made variable name clearer

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -19,6 +19,44 @@
...
+// Number of bytes in a pixel
+const int bytesPerPixel = 4;
+struct pixelColorARGB {
```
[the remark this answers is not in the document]
jeremyrhyde: I went back and forth on this name, but will go with ColorARGB unless anyone has a different preference

viam-cartographer/src/slam_service/slam_service.cc
jeremyrhyde: moving all variables non consts to snake case due to google C++ style guidelines

viam-cartographer/src/slam_service/slam_service.cc
jeremyrhyde: moving all variables non consts to snake case due to google C++ style guidelines

viam-cartographer/src/slam_service/slam_service.cc
jeremyrhyde: Corrected comment and uppercase function name

viam-cartographer/src/slam_service/slam_service.cc
jeremyrhyde: added new line :)

nicksanford:
High-level questions before PR review:
1. Do we need to be confident we are not adding latency to GetPointCloudMap response time?
2. If yes, do we know how this change affects GetPointCloudMap response times?

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -19,6 +19,44 @@
...
+    return (pixel_color.R == defaultNoDataRGBValue) &&
+           (pixel_color.G == defaultNoDataRGBValue) &&
+           (pixel_color.B == defaultNoDataRGBValue);
+}
+
+// Conver the scale of a specified pixel color channel from the given
+// 102-255 range to 100-0. This is an initial solution for extracting
+// probability information from cartographer
+int CalculateProbabilityFromColorChannels(ColorARGB pixel_color) {
+    unsigned char max_val = CHAR_MAX;
```
nicksanford, line 47: This is equal to 127.

I think you might want `UCHAR_MAX`?

Also, if it is important that it be from 102-255... maybe this can just be another constant like `maxDataRGBValue` which we set to 255?

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -19,6 +19,44 @@
...
+    unsigned char min_val = defaultNoDataRGBValue;
+    unsigned char max_prob = 100;
+    unsigned char min_prob = 0;
+
+    // Probability is currently determined solely by the R channel
+    unsigned char color_channel_val = pixel_color.R;
+    unsigned char prob = (max_val - color_channel_val) * (max_prob - min_prob) /
```
nicksanford, line 54: Are we certain `color_channel_val` can't be lower than 102?

If we are then I'm good with leaving this as is. 

If we are not certain but want to represent any color channel value equal to, or below 102 as probability 0, I recommend doing the following change to prevent returning probabilities outside of 0 - 100
```suggestion
    unsigned char prob = (max_val - std::max(color_channel_val, min_val)) * (max_prob - min_prob) /
```

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -333,97 +371,82 @@ void SLAMServiceImpl::GetLatestSampledPointCloudMapString(
...
     }
 
+    // Get data from painted surface in ARGB32 format
     auto painted_surface = painted_slices->surface.get();
+    auto image_format = cairo_image_surface_get_format(painted_surface);
+    if (image_format != cartographer::io::kCairoFormat) {
+        std::string error_log =
+            "Error cairo surface in wrong format, expected Cairo_Format_ARGB32";
+        LOG(ERROR) << error_log;
+        throw std::runtime_error(error_log);
```
nicksanford, line 381: [q] Does this terminate the cartogropher grpc server?

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -333,97 +371,82 @@ void SLAMServiceImpl::GetLatestSampledPointCloudMapString(
         }
     }
 
+    // Get data from painted surface in ARGB32 format
     auto painted_surface = painted_slices->surface.get();
+    auto image_format = cairo_image_surface_get_format(painted_surface);
+    if (image_format != cartographer::io::kCairoFormat) {
+        std::string error_log =
+            "Error cairo surface in wrong format, expected Cairo_Format_ARGB32";
```
nicksanford, line 379: Could we please log what format we did get?

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -333,97 +371,82 @@ void SLAMServiceImpl::GetLatestSampledPointCloudMapString(
...
+        LOG(ERROR) << error_log;
```
nicksanford, line 380:
```suggestion
        LOG(ERROR) << error_log << " got " <<  image_format;
```

nicksanford (changes requested):

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -333,97 +371,82 @@ void SLAMServiceImpl::GetLatestSampledPointCloudMapString(
...
+    std::string pcd_data;
+    for (int pixel_y = 0; pixel_y < height; pixel_y++) {
+        for (int pixel_x = 0; pixel_x < width; pixel_x++) {
+            // Get byte index associated with pixel
+            int pixel_index = pixel_x + pixel_y * width;
+            int byte_index = pixel_index * bytesPerPixel;
+
+            // We assume we are running on a little-endian system, so the ARGB
+            // order is reversed
+            ColorARGB pixel_color;
```
nicksanford, line 402: @jeremyrhyde this is so much easier to follow!!!

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -333,97 +371,82 @@ void SLAMServiceImpl::GetLatestSampledPointCloudMapString(
...
-        viam::utils::writeFloatToBufferInBytes(data_buffer,
-                                               rotated_map_point.x());
-        viam::utils::writeFloatToBufferInBytes(data_buffer,
-                                               rotated_map_point.y());
-        viam::utils::writeFloatToBufferInBytes(data_buffer,
-                                               rotated_map_point.z());
-        viam::utils::writeIntToBufferInBytes(data_buffer, prob);
+            // Convert pixel location to pointcloud point in meters
+            float x_pos = (pixel_x - origin_pixel_x) * resolutionMeters;
+            // Y is inverted to match output from getPosition()
```
nicksanford, line 422: [q] I know this is how the code was before but why do we need to invert Y in GetPosition()?

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -333,97 +371,82 @@ void SLAMServiceImpl::GetLatestSampledPointCloudMapString(
...
-    int size_data = width * height * 4;
-
-    // Each pixel contains 4 bytes of information in RGBA format
-    // data_vect[i + 0] is the R channel
-    // data_vect[i + 1] is the B channel
-    // data_vect[i + 2] is the G channel
-    // data_vect[i + 3] is the A channel
-    std::vector<unsigned char> data_vect(data, data + size_data);
+    // Get pixel containing map origin (0, 0)
+    float origin_pixel_x = painted_slices->origin.x();
```
nicksanford, line 388: [q] will `origin_pixel_x` and origin_pixel_y` ever be non 0?

viam-cartographer/src/slam_service/slam_service.h
```diff
@@ -37,24 +38,22 @@ namespace viam {
...
 // Quaternion to rotate axes to the XZ plane
 static const Eigen::Quaterniond pcdRotation(0.7071068, -0.7071068, 0, 0);
 // Static offset quaternion, so orientation matches physical intuition.
 // This will result in rotations occurring within the y axis to match 2D mapping
 // in the XZ plane
 static const Eigen::Quaterniond pcdOffsetRotation(0.7071068, 0.7071068, 0, 0);
+// The resolutionMeters variable defines the area in meters that each pixel
+// represents. This is used to draw the jpeg map and in so doing defines the
+// resolution of the outputted PCD
+static const double resolutionMeters = 0.05;
```
nicksanford, line 55: So this means we output at 5cm resolution?

viam-cartographer/src/slam_service/slam_service.cc
nicksanford: makes sense thanks!

viam-cartographer/src/slam_service/slam_service.cc
nicksanford: The simplified version you have now made it easy to follow thanks!

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -333,97 +371,82 @@ void SLAMServiceImpl::GetLatestSampledPointCloudMapString(
...
 
-        viam::utils::writeFloatToBufferInBytes(data_buffer,
-                                               rotated_map_point.x());
-        viam::utils::writeFloatToBufferInBytes(data_buffer,
-                                               rotated_map_point.y());
-        viam::utils::writeFloatToBufferInBytes(data_buffer,
-                                               rotated_map_point.z());
-        viam::utils::writeIntToBufferInBytes(data_buffer, prob);
+            // Convert pixel location to pointcloud point in meters
+            float x_pos = (pixel_x - origin_pixel_x) * resolutionMeters;
```
nicksanford, line 421: [q] One more (maybe silly question) but I don't understand how this reduces the resolution of the map.

I would expect that `reducing the resolution` involves representing multiple points within the resolution (5cm) as a single point with some aggregate of the probabilities of all points that were within that 2d area. 

It is not clear to me how multiplying every point with a probability by `0.05` does that.

1. Aren't we still returning the same number of points?
2. Where are we aggregating the probabilities of multiple points within the same resolution area?

nicksanford: We synced offline:
The answer is that `viam::io::PaintSubmapSlices` takes the resolution & calls cairo with it to scale the painted map. Cairo then does points 1 & 2. We need to multiply by resolution on this line to convert from pixels back into meters.

jeremyrhyde: Resolved offline. The resolution comes into effect during the painting. This use of it here is jsut for converting back to realworld space.

X (meters) = Y (pixels) * resolution (meter/pixel)

viam-cartographer/src/slam_service/slam_service.cc
jeremyrhyde: Yes, this tracks the origin location of the robot and as the images grows in all dimensions as you add submaps, this values can shift to non 0,0.

viam-cartographer/src/slam_service/slam_service.cc
jeremyrhyde: This is non-trivial as it is a custom type that when cast to string just appears as 0,1,2,3,4... In the future we could add a utility function to convert this into a usage string but since this can not be affected by us or our users and only by fundamental changes to cartographer we will not pursue this at this point.

viam-cartographer/src/slam_service/slam_service.cc
jeremyrhyde: see comment below

viam-cartographer/src/slam_service/slam_service.cc
jeremyrhyde: Yes. This function appears 3 times in the code:
- GRPC call to GetPointCloudMap - in a try/catch statement which will stop cartographer if an error occurs via terminate
- CacheMapInLocalizationMode - try/catch statement which will stop cartographer via terminate
- BackupLatestMap - no try/catch.

Question to the team, should this terminate cartographer? And if not should we add a try/catch for BackupLatestMap

viam-cartographer/src/slam_service/slam_service.cc
jeremyrhyde: This is correct, I checked the max value in a channel for a large map run and recorded 253. Changing this to UCHAR_MAX

viam-cartographer/src/slam_service/slam_service.cc
jeremyrhyde: @JohnN193 question for you John. I recorded the minimum value for pixel and it was 7 not 102. Does this need to be changed? Or is there a reason we would want to threshold it at 102.

jeremyrhyde:
> High-level questions before PR review:
> 
> 1. Do we need to be confident we are not adding latency to GetPointCloudMap response time?
> 2. If yes, do we know how this change affects GetPointCloudMap response times?

Hi Nick, I am planning on running QA tests of this to ensure the session_expired issue does not get exacerbated by it. That being said an initial test comparing the final map of our old viam office produced by the old method resulted in a 9.6MB PCD, with this new method it was only 3.8MB. So we should see some improvements to the session_expired issue when mapping areas of that size or smaller

This resolution-based mapping will cause issues for large maps as it is unbounded, however, that will be up to the user once we move to a run-time-based config parameter for this. For now, this default value can work well on relatively large areas and is dense enough to satisfy the need of motion.

viam-cartographer/src/slam_service/slam_service.cc
JohnN193: If you are seeing the probability channel register numbers below 102 consistently then feel free to change.

viam-cartographer/src/slam_service/slam_service.cc
JohnN193: If we think the format will also fail in future GetPointCloudMap calls if it triggers once then this should terminate cartographer

viam-cartographer/src/main.cc
```diff
@@ -53,7 +53,14 @@ int main(int argc, char** argv) {
     // This log line is needed by rdk to get the port.
     LOG(INFO) << "Server listening on " << *selected_port << "\n";
 
-    slamService.RunSLAM();
+    try {
+        slamService.RunSLAM();
```
nicksanford, line 57: Note: This is being added so that the program terminates in all cases if https://github.com/viamrobotics/viam-cartographer/pull/25/files#diff-92ff8a5cd3a57430d78c6ba3b61a91d1810399b77613eece3c55841868d2822eR384 throws.

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -19,6 +19,47 @@
...
+// 102-255 range to 100-0. This is an initial solution for extracting
+// probability information from cartographer
+int CalculateProbabilityFromColorChannels(ColorARGB pixel_color) {
+    unsigned char max_val = UCHAR_MAX;
+    unsigned char min_val = defaultNoDataRGBValue;
+    unsigned char max_prob = 100;
+    unsigned char min_prob = 0;
+
+    // Probability is currently determined solely by the R channel
+    // Values are restricted to be above defaultNoDataRGBValue to ensure
```
nicksanford, line 53:
```suggestion
    // Values are restricted to be above or equal to defaultNoDataRGBValue to ensure
```

viam-cartographer/src/main.cc
jeremyrhyde: *from the main thread

nicksanford (approved):

PR #31, opened 2023-03-27
fix-race-test-race-condition

nicksanford:

PR #32, opened 2023-03-28
revert-setupTestGRPCServer-to-private

nicksanford:
Attempts to fix: https://github.com/viamrobotics/viam-cartographer/actions/runs/4536705514/jobs/7993714034 

Currently, tests are flakey as sometimes the server returns an error. Given that modularization v2 will remove the grpc server, this change doesn't investigate the cause of the error & instead reverts the code back to the pre modularization v1 code state which can still be seen in orbslam here: https://github.com/viamrobotics/viam-orb-slam3/blob/main/viam-orb-slam3_test.go#L77

### DIFF
--- a/viam-cartographer_test.go
+++ b/viam-cartographer_test.go
@@ -8,6 +8,7 @@ package viamcartographer_test
 import (
 	"context"
 	"fmt"
+	"net"
 	"os"
 	"strconv"
 	"testing"
@@ -19,6 +20,7 @@ import (
 	slamTesthelper "go.viam.com/slam/testhelper"
 	"go.viam.com/test"
 	"go.viam.com/utils"
+	"google.golang.org/grpc"
 
 	"github.com/viamrobotics/viam-cartographer/internal/testhelper"
 )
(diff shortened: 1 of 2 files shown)


viam-cartographer_test.go
```diff
@@ -336,3 +338,15 @@ func TestSLAMProcess(t *testing.T) {
 
 	testhelper.ClearDirectory(t, dataDir)
 }
+
+// SetupTestGRPCServer sets up and starts a grpc server.
```
kkufieta, line 342:
```suggestion
// setupTestGRPCServer sets up and starts a grpc server.
```

kkufieta (approved):
LGTM once linter passes (I think you simply have to remove `logger` from `setupTestGRPCServer`)

viam-cartographer_test.go
```diff
@@ -336,3 +338,14 @@ func TestSLAMProcess(t *testing.T) {
...
+// It returns the grpc server and the port at which it is served.
+func setupTestGRPCServer(tb testing.TB) (*grpc.Server, int) {
```
EshaMaharishi, line 344: we had to move this back into the same package in order to not handle the error return value of grpcServer.Serve. see also https://github.com/viamrobotics/viam-cartographer/pull/31/files#r1151106610

the grpc server being set up here is a mock that simulates carto_grpc_server

EshaMaharishi (approved):

PR #34, opened 2023-03-29
RSDK-2016 - Add module

kkufieta:
Ticket: https://viam.atlassian.net/browse/RSDK-2016

* Implements module

Tested on:
* MacOS Intel
* RPI


module/main.go
```diff
@@ -1,2 +1,38 @@
 // Package main is a module with a cartographer SLAM service model.
 package main
+
+import (
+	"context"
+
+	"github.com/edaniels/golog"
+	viamcartographer "github.com/viamrobotics/viam-cartographer"
```
EshaMaharishi, line 8: For my knowledge, do you know why we need to alias this to `viamcartographer`, but didn't need add an alias for `golog`?

viam-cartographer.go
```diff
@@ -133,7 +133,8 @@ func New(
 		return nil, err
 	}
 
-	cancelCtx, cancelFunc := context.WithCancel(ctx)
+	// Need to pass in a long-lived context because ctx is short-lived
+	cancelCtx, cancelFunc := context.WithCancel(context.Background())
```
EshaMaharishi, line 137: Can we talk about this change in person? I know this was one of the issues you and Zack were looking into, but I don't fully understand it.

EshaMaharishi (commented):
Looks good, but I have a question

module/main.go
zaporter-work: I think the answer to your question is: 
`viam-cartographer` has a hyphen, go package names must be identifiers, identifiers cannot contain hyphens. golog does not have a hypen

EshaMaharishi: Ah! Thanks.

viam-cartographer.go
zaporter-work: Yes, we should talk about it with Nick as well. You might want to come to our meeting in C8 at 15:00 for the first 10 minutes so we can talk about this.

EshaMaharishi: Sounds great, thank you!

zaporter-work:
I think it makes sense to setup an example config and switch to TransformMapAttributes...  in this PR (similar to orb). Happy to approve if you want to split though

zaporter-work (approved):

module/sample_linux.json
```diff
@@ -0,0 +1,44 @@
...
+        }
+      }
+    ],
+    "modules": [
+      {
+        "executable_path": "/usr/local/bin/rplidar-module",
+        "name": "rplidar_module"
+      },
+      {
+        "executable_path": "/Users/kat/viam/viam-cartographer/bin/cartographer-module",
```
zaporter-work, line 40: [nit] Why isnt this /usr/local/bin like rplidar_module?

module/sample_linux.json
```diff
@@ -0,0 +1,44 @@
...
+      }
+    ],
+    "services": [
+      {
+        "name": "slam-test",
+        "model": "viam:slam:cartographer",
+        "type": "slam",
+        "attributes": {
+          "port": "localhost:8083",
+          "data_dir": "/Users/kat/viam/datasets/test-delete-me",
```
zaporter-work, line 18: [q] should we hardcode your name in the sample config? Why dont we do something like `HOME_DIRECTORY/DATA_DIR` to indicate to the user that they should use replace this value?

viam-cartographer.go
```diff
@@ -133,7 +128,8 @@ func New(
 		return nil, err
 	}
 
-	cancelCtx, cancelFunc := context.WithCancel(ctx)
+	// Need to pass in a long-lived context because ctx is short-lived
+	cancelCtx, cancelFunc := context.WithCancel(context.TODO())
```
nicksanford, line 132: I verified that the orb module os process terminates (either immediately or after a few seconds) regardless of whether viam-server is killed via `SIGTERM` `SIGQUIT` or `SIGKILL`.

nicksanford: sry, I posted on the wrong PR. Ignore this comment

viam-cartographer.go
```diff
@@ -133,7 +128,8 @@ func New(
 		return nil, err
 	}
 
-	cancelCtx, cancelFunc := context.WithCancel(ctx)
+	// Need to pass in a long-lived context because ctx is short-lived
```
nicksanford, line 131:
```suggestion
	// Need to pass in a long-lived context because ctx is short-lived
	// To be re-evaluated in https://viam.atlassian.net/browse/RSDK-2586
```

viam-cartographer.go
nicksanford: I changed this to context.Background() to be in line with orbslam based on recent learnings: https://github.com/viamrobotics/viam-orb-slam3/pull/29/files#diff-c70dd5ed993c1b53cd893d5a38fa4665c771456384a369b2a67b7a55ddb9fffeR341

EshaMaharishi: Thank you! Resolving

module/sample_linux.json
nicksanford: updated

module/sample_linux.json
nicksanford: updated

nicksanford (approved):

module/sample_linux.json
```diff
@@ -0,0 +1,44 @@
+{
+  "components": [
+    {
+      "model": "viam:lidar:rplidar",
```
EshaMaharishi, line 4: [note to self] if you're referencing a modular resource, you need to give it the full name like we do here. if you're referencing a builtin resource, you give just the short name like we do [here](https://github.com/viamrobotics/viam-orb-slam3/pull/29/files#diff-868c9c2fdc104b26b1d16968030f23a5b552546a5bb86fbda780057b847cc41aR10).

i'm not sure if 'namespace' (L6 below) is related in any way.

EshaMaharishi (approved):

PR #42, opened 2023-04-06
RSDK-2546 Fix Timestamp Flaky Test

jeremyrhyde:
Current testing of the filename timestamp generation is flaky due to the precision level of chrono's time request function. This has been fixed by requiring a timestamp to be sent to the filename generation function thus making it time independent. 

For the test, this means that instead of having to record the time before and after the filename generation step and confirm the timestamp is in between these two values, we are able to directly check if the timestamp parsed from the filename matches the one inputted into the function.

Testing was done by running the unit test 10,000 times. Zero failures were seen as compared to ~5 failures seen using the older method.

Note: an alternative path could be to add millisecond resolution to the filename generation, however, making this function time-independent is more correct without changing the underlying behavior of the current setup. We can still choose to pursue millisecond accuracy in the future should it be desired.

JIRA Ticket: https://viam.atlassian.net/browse/RSDK-2546

### DIFF
--- a/viam-cartographer/src/io/file_handler.cc
+++ b/viam-cartographer/src/io/file_handler.cc
@@ -19,8 +19,9 @@ namespace io {
 
 namespace fs = boost::filesystem;
 
-const std::string MakeFilenameWithTimestamp(std::string path_to_dir, std::time_t t) {
-    //std::time_t t = std::time(nullptr);
+const std::string MakeFilenameWithTimestamp(std::string path_to_dir,
+                                            std::time_t t) {
+    // std::time_t t = std::time(nullptr);
     char timestamp[100];
     std::strftime(timestamp, sizeof(timestamp), time_format.c_str(),
                   std::gmtime(&t));
--- a/viam-cartographer/src/io/file_handler.cc
+++ b/viam-cartographer/src/io/file_handler.cc
@@ -19,8 +19,8 @@ namespace io {
 
 namespace fs = boost::filesystem;
 
-const std::string MakeFilenameWithTimestamp(std::string path_to_dir) {
-    std::time_t t = std::time(nullptr);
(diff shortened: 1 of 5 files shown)


viam-cartographer/src/io/file_handler.cc
```diff
@@ -19,8 +19,9 @@ namespace io {
 
 namespace fs = boost::filesystem;
 
-const std::string MakeFilenameWithTimestamp(std::string path_to_dir) {
-    std::time_t t = std::time(nullptr);
+const std::string MakeFilenameWithTimestamp(std::string path_to_dir,
+                                            std::time_t t) {
+    // std::time_t t = std::time(nullptr);
```
JohnN193, line 24: [nit] remove

JohnN193 (approved):
LGTM

viam-cartographer/src/io/file_handler.cc
nicksanford: yep, please remove

viam-cartographer/src/slam_service/slam_service.h
```diff
@@ -7,6 +7,7 @@
 #include <grpcpp/server_context.h>
 
 #include <atomic>
+#include <ctime>
```
nicksanford, line 10: why is this needed?

nicksanford (changes requested):

viam-cartographer/src/slam_service/slam_service.h
jeremyrhyde: We now are calling std::time on line 660 of slam_service.cc. however, since we import file_handler.h which also imports this

EshaMaharishi (approved):
Good solution!

nicksanford (approved):

PR #50, opened 2023-04-18
[RSDK-2760] Setup script should force install openssl

dmhilly:
When I tried to build cartographer on my Mac for the first time, I ran into issues with OpenSSL (sadly I don't remember what the error was and didn't save it). @nicksanford and I cross-referenced the `setup_cartographer_macos` script with the `setup_orbslam_macos` script and observed that in orb slam, we run `brew link openssl` with the `--force` flag. When I ran this locally, I stopped getting errors. This PR changes the `setup_cartographer_macos` script to run the correct command.


nicksanford (commented):
will approve as soon as it has a ticket associated with it

nicksanford (approved):

PR #52, opened 2023-04-19
[RSDK-2560] Don't update map in localization mode

dmhilly:
This PR fixes a bug where we were updating the map even when in localization mode. I fixed this by ensuring we're not in localization mode before pulling the latest map from cartographer and updating our cached map. 

@jeremyrhyde pointed out that even with this fix, the cartographer algorithm will still be updating its internal implementation of the map when we're in localization mode, we just won't ask it for that updated map. A better long-term solution might be to tell it to stop updating the map when we're localizing. For now, I've implemented the simplest fix required to address the bug.

### Manual Testing

I tested online SLAM on my Mac with an rplidar, using the configuration below. 
<details>
   <summary> config.json </summary>

```json
{
  "modules": [
    {
      "executable_path": "/opt/homebrew/bin/rplidar-module",
      "name": "rplidar_module"
    }
  ],
  "components": [
    {
      "attributes": {
        "device_path": "/dev/tty.usbserial-0001"
      },
      "name": "rplidar",
      "namespace": "rdk",
      "type": "camera",
      "depends_on": [],
      "model": "viam:lidar:rplidar",
      "service_config": [
        {
          "type": "data_manager",
          "attributes": {
            "capture_methods": [
              {
                "disabled": false,
                "method": ""
              }
            ]
          }
        }
      ]
    }
  ],
  "services": [
    {
      "attributes": {
        "data_dir": "/Users/devinhilly/cartographer_dir",
        "map_rate_sec": 0,
        "data_rate_msec": 200,
        "delete_processed_data": false,
        "use_live_data": true,
        "sensors": [
          "rplidar"
        ],
        "config_params": {
          "min_range": "0.3",
          "max_range": "12",
          "mode": "2d"
        }
      },
      "model": "cartographer",
      "name": "slam-svc",
      "type": "slam"
    }
  ]
}
```
</details>

I first ran SLAM with map_rate_sec = 60 to build the map, then ran it with map_rate_sec = 0 to trigger localization mode. I confirmed that SLAM started in localization mode using the logs. Without the fix, you can see it updates the map in the video below. Both videos are about 1 minute long, but I played them back at a faster speed for you all to view more easily:

https://user-images.githubusercontent.com/9754276/233167815-910cce02-36db-4d50-84eb-e7c8ddf477e7.mov

With the fix, it doesn't update the map:

https://user-images.githubusercontent.com/9754276/233167406-19905a17-1450-42a0-a8fb-b081d14f12b4.mov


EshaMaharishi (approved):

jeremyrhyde (approved):
LGTM. As Devin pointed out the updating this issue that this PR address is still going on in the cartographer. however, we currently do not know if this is a real issue or a by product of the way we produce a pointcloud. 

Currently pointcloud production uses cairo to paint each and ever submap. A localization run DOES produce new submaps however, they **may** (this is not known for certain) not be used during localization mode except by cairo when we iterate through them all. We could consider tackling this by only painting submaps that are connected to trajectory ids 0 -> n-1, but since we would still like to cache the map to avoid the high compute process involved with creating it, this is a moot point in my opinion. 

If desired, we can look more deeply into the behavior of cartographer during localization mode to determine if these submaps creating during localization are being used incorrectly.

nicksanford (approved):

PR #59, opened 2023-04-24
RSDK-2823 Update Cartographer Output to be on XY Plane

jeremyrhyde:
This PR removes the previous post-processing that was being performed on Cartographer to rotate it onto the XZ plane. This also updates the integration tests for the new position output

JIRA Ticket: https://viam.atlassian.net/browse/RSDK-2823

Associated PR: 
- https://github.com/viamrobotics/rdk/pull/2269
- https://github.com/viamrobotics/api/pull/251


integration_test.go
```diff
@@ -43,14 +43,14 @@ func testCartographerMap(t *testing.T, svc slam.Service) {
 
 	pointcloud, _ := pointcloud.ReadPCD(bytes.NewReader(pcd))
 	t.Logf("Pointcloud points: %v", pointcloud.Size())
-	test.That(t, pointcloud.Size(), test.ShouldBeGreaterThanOrEqualTo, 100)
+	test.That(t, pointcloud.Size(), test.ShouldBeGreaterThanOrEqualTo, 45)
```
jeremyrhyde, line 46: [placeholder comment] 

```
    integration_test.go:61: Position point: (-0.003358475751032404, 0.005523531576245688, 0)
    integration_test.go:67: Position orientation: RX: 0, RY: 0, RZ: -1, Theta: 0.0014163465941501687
    integration_test.go:45: Pointcloud points: 45
    integration_test.go:46: Expected '45' to be greater than or equal to '100' (but it wasn't)!
--- FAIL: TestCartographerIntegration2D (3.30s)
```

jeremyrhyde, line 46: [note] unclear as to why this is changing. Doing additional research into the reasons behind it but may want to submit the PR in advance to unblock motion

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -396,15 +398,9 @@ void SLAMServiceImpl::GetLatestSampledPointCloudMapString(
...
-            Eigen::Vector3d rotated_map_point = pcdRotation * map_point;
-
-            viam::utils::writeFloatToBufferInBytes(pcd_data,
-                                                   rotated_map_point.x());
-            viam::utils::writeFloatToBufferInBytes(pcd_data,
-                                                   rotated_map_point.y());
-            viam::utils::writeFloatToBufferInBytes(pcd_data,
-                                                   rotated_map_point.z());
+            viam::utils::writeFloatToBufferInBytes(pcd_data, x_pos);
+            viam::utils::writeFloatToBufferInBytes(pcd_data, x_pos);
```
JohnN193, line 402: x_pos is written twice

integration_test.go
```diff
@@ -43,14 +43,14 @@ func testCartographerMap(t *testing.T, svc slam.Service) {
...
 }
 
 // Checks the cartographer position within a defined tolerance.
 func testCartographerPosition(t *testing.T, svc slam.Service, expectedComponentRef string) {
-	expectedPos := r3.Vector{X: -0.004, Y: 0, Z: -0.004}
+	expectedPos := r3.Vector{X: -0.004, Y: -0.004, Z: 0}
 	tolerancePos := 0.04
-	expectedOri := &spatialmath.R4AA{Theta: 0, RX: 0, RY: 1, RZ: 0}
+	expectedOri := &spatialmath.R4AA{Theta: 0, RX: 0, RY: 0, RZ: -1}
```
JohnN193, line 53: well i guess we found why the map is "flipped"

JohnN193 (approved):
Saw a typo but otherwise LGTM

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -74,24 +74,26 @@ ::grpc::Status SLAMServiceImpl::GetPosition(ServerContext *context,
 
     // Rotate pose to XZ plane. Additional angle offset is used so rotations
```
nicksanford, line 75: 1. Can we please remove these comments and grep the codebase for any other mention of the XZ plane & remove them?
2. Can we please remove `pcdRotation` if it is no longer used?
3.  Can we please remove `pcdOffsetRotation` if it is no longer used?

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -396,15 +398,9 @@ void SLAMServiceImpl::GetLatestSampledPointCloudMapString(
...
-            Eigen::Vector3d map_point(x_pos, y_pos, z_pos);
-            Eigen::Vector3d rotated_map_point = pcdRotation * map_point;
-
-            viam::utils::writeFloatToBufferInBytes(pcd_data,
-                                                   rotated_map_point.x());
-            viam::utils::writeFloatToBufferInBytes(pcd_data,
-                                                   rotated_map_point.y());
-            viam::utils::writeFloatToBufferInBytes(pcd_data,
-                                                   rotated_map_point.z());
+            viam::utils::writeFloatToBufferInBytes(pcd_data, x_pos);
```
nicksanford, line 401: Why are we still flipping the sign of y_pos?

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -74,24 +74,26 @@ ::grpc::Status SLAMServiceImpl::GetPosition(ServerContext *context,
...
+    // pcdRotation;
+    auto pos_vector = global_pose.translation();
+    auto pos_quat = global_pose.rotation();
 
     // Set pose for our response
     Pose *myPose = response->mutable_pose();
-    myPose->set_x(rotated_vector.x());
-    myPose->set_y(rotated_vector.y());
-    myPose->set_z(rotated_vector.z());
+    myPose->set_x(pos_vector.x());
```
nicksanford, line 85: Are we sure that nothing needs to change on the `GetPosition` endpoint?

integration_test.go
nicksanford: why?

nicksanford (changes requested):

integration_test.go
JohnN193: cartographer is writing the orientation with Z down(at least that's how I am reading it). I don't think the previous transforms took that into account

viam-cartographer/src/slam_service/slam_service.cc
jeremyrhyde: I have chosen not to deal with the flipping issue in this PR as I cannot validate the quaternion without the SLAM UI update or some substantial work messing around with quaternions. Therefore the negative y_pos was left unaffected by this PR

viam-cartographer/src/slam_service/slam_service.cc
jeremyrhyde: Nice catch! That resolves the issue around # of map points decreasing in the integration test!

viam-cartographer/src/slam_service/slam_service.cc
jeremyrhyde: Removed all comments and all association with XZ/rotations

integration_test.go
jeremyrhyde: Would it be more accurate to say, we've now found a symptom of the map flip? Root cause is still TBD (although probably has to do with the -y we set in the map

viam-cartographer/src/slam_service/slam_service.cc
jeremyrhyde: Rotation of the position was removed on lines 77-78

integration_test.go
JohnN193: if the -y gets removed then the position has to be rotated accordingly

JohnN193: but this is a convo for another PR

viam-cartographer/src/slam_service/slam_service.cc
nicksanford: got it, sounds good

PR #84, opened 2023-05-04
Fix Readme for seamless setup & build process

kkufieta:
I ran into issues with running cartographer today after uninstalling and attempting to reinstall everything. What was missing was running the `make install-lua-files` command, which is why I think we should add it to the set of instructions in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,10 +88,11 @@ These containers are set to persist between sessions via the `persistent` parame
 make bufinstall && make buf 
 # Install dependencies
 make setup
+# Build the binary
+make build
 # Install lua files
 make install-lua-files
-# Build & install the binary
-make build
+# Install the cartographer-module binary
 make install
 # Run the binary
 cartographer-module
--- a/README.md
+++ b/README.md
@@ -88,6 +88,8 @@ These containers are set to persist between sessions via the `persistent` parame
 make bufinstall && make buf 
 # Install dependencies
 make setup
+# Install lua files
+make install-lua-files
 # Build & install the binary
 make build
 make install


README.md
```diff
@@ -88,6 +88,8 @@ These containers are set to persist between sessions via the `persistent` parame
 make bufinstall && make buf 
 # Install dependencies
 make setup
+# Install lua files
+make install-lua-files
```
EshaMaharishi, line 92: Should we do this step after `make build` to match what we do in the appimage workflow (https://github.com/EshaMaharishi/viam-cartographer/blob/bbeb5194514f988d15b14bc336578e370ec48154/.github/workflows/appimage.yml#L64)?

Or alternatively should we make the appimage workflow follow the order here?

EshaMaharishi (approved):
LGTM mod comment

README.md
kkufieta: good idea - will align it with how the appimage workflow looks

README.md
```diff
@@ -88,8 +88,11 @@ These containers are set to persist between sessions via the `persistent` parame
 make bufinstall && make buf
```
nicksanford, line 88: Why do we need to run `make buf` at this step.

The order I would have expected would be:
#Install brew / apt packages. Only needs to be done once.
make setup

#install buf. Only needs to be done once 
make bufinstall

#copy lua /config files into /usr/local. Only needs to be done once. Doesn't change during compilation:
make install-lua-files

#needs to be run after `make clean`
make buf

#Install the cartographer-module binary
make build

#Install the cartographer-module binaries 
make install

#Run the binary
cartographer-module

nicksanford (changes requested):

nicksanford (approved):
Approved. The comments I brought up will be integrated in a later PR.

PR #92, opened 2023-05-16
RSDK-2881: Update viam-cartographer to check if camera provides timestamps

dmhilly:
### Summary

This PR modifies the viam-cartographer to provide metadata in the context it provides when calling its sensor. If the sensor is a replay camera, the camera will attach metadata to the gRPC response which will then get injected into the context in a client-side gRPC interceptor. When viam-cartographer finds timestamps in the context metadata, it extracts the time received value and uses this and the timestamp for the PCD file name.

### Manual Testing

I tested by configuring a robot (my Mac) with a replay camera and providing that as a sensor for SLAM. Along with the changes in the corresponding [rdk PR](https://github.com/viamrobotics/rdk/pull/2371), I confirmed that the files get written to disk with the correct timestamps correlated with the time they were captured (last week):

Screenshot 2023-05-16 at 3 31 18 PM


internal/dim-2d/dim-2d.go
```diff
@@ -105,6 +107,7 @@ func GetAndSaveData(ctx context.Context, dataDirectory string, lidar lidar.Lidar
 	ctx, span := trace.StartSpan(ctx, "viamcartographer::internal::dim2d::GetAndSaveData")
 	defer span.End()
 
+	ctx, md := utils.ContextWithMetadata(ctx)
```
jeremyrhyde, line 110: Based on the results of the discussion regarding my open question, update the comment here to more correctly represent that the deciding factor is the parsing/extraction and not the presence of the timestamp.

Example:
```suggestion
	// If the timestamps correlated with the point cloud and provided by the server can be extracted use them when writing the data to disk, otherwise use the current time.
```

jeremyrhyde (changes requested):
[q] The extracting of the timestamp here is being performing by accessing the TimeReceivedMetadataKey and parsing the timestamp out of it. That leaves me with two questions:

1. If the camera's context metadata does not contain that key, won't ok return `false`? Ideally, that should be the check to determine if a camera is in replay or not. 
2. The current implementation uses the result of the **Parsing** of the timeReceived to determine if a given camera is a replay camera or not. This mean if a replay camera ever had malformed metadata, it would register as a real/live camera. 

Please correct me if those questions are overlooking something, However, I dont believe we should be using the parsing to determine replay status, only the presence of that `camera.TimeReceivedMetadataKey`

dmhilly:
@jeremyrhyde We should never get metadata for that field in the wrong format, but anything can happen :) My thought was it would be better to just use current timestamps (`time.Now()`s) if that field is set and malformed, rather than writing the malformed value to disk and having cartographer get confused. We could also just error out; maybe that would be better.

edaniels (approved):
nice, lgtm!

PR #95, opened 2023-05-22
RSDK-3135 - create modularization v2 enabled feature flag

kim-mishra:
### Changes
- add in feature flag to config
- handle retrieving feature flag in `GetOptionalParameters`

### Testing
- cartographer works with no feature flag provided
- cartographer works with the feature flag provided and set to true
- cartographer works with the feature flag provided and set to false


config/config.go
```diff
@@ -109,10 +110,17 @@ func GetOptionalParameters(config *Config, defaultPort string,
...
-		return "", 0, 0, false, false, err
+		return "", 0, 0, false, false, false, err
 	}
 
 	deleteProcessedData := DetermineDeleteProcessedData(logger, config.DeleteProcessedData, useLiveData)
 
-	return port, dataRateMsec, mapRateSec, useLiveData, deleteProcessedData, nil
+	modularizationV2Enabled := false
+	if config.ModularizationV2Enabled == nil {
+		logger.Debug("no modularization_v2_enabled given, continuing with modularization v1")
```
EshaMaharishi, line 120: We should leave this debug message out (though I believe it won't show up anyway due to https://viam.atlassian.net/browse/RSDK-2842), since it probably doesn't make sense to mention a feature flag to production users when the feature is still under development and not turned on. If you want you could log when the flag _is_ enabled, for us, though.

EshaMaharishi (approved):
LGTM mod comment!

Since Kat is OOO today and might be OOO tomorrow, let's move Kat to assignee so that she's aware but go ahead and merge.

PR #102, opened 2023-05-26
RSDK-3323 - addSensorReading refactor

kim-mishra:
### Changes
Move parts of code in `ProcessDataAndStartSavingMaps` that are not file syste dependent into new method called `AddSensorReading`

### Testing
- Viewed cartographer map on main and then again on this branch and they looked the same.


viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -647,6 +647,23 @@ void SLAMServiceImpl::SaveMapWithTimestamp() {
...
+
+    if (measurement.ranges.size() > 0) {
+        trajectory_builder->AddSensorData(kRangeSensorId.id, measurement);
+        auto local_poses = map_builder.GetLocalSlamResultPoses();
+        if (local_poses.size() > 0) {
+            tmp_global_pose =
+                map_builder.GetGlobalPose(trajectory_id, local_poses.back());
+        }
+    }
+    return tmp_global_pose;
```
kkufieta, line 664: [nit] This function should ideally return `void`, and `latest_global_pose` should be set within this function since updating it goes hand in hand with adding a new sensor reading.

kkufieta (changes requested):

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -647,6 +647,23 @@ void SLAMServiceImpl::SaveMapWithTimestamp() {
     }
 }
 
+cartographer::transform::Rigid3d SLAMServiceImpl::AddSensorReading(
+    cartographer::sensor::TimedPointCloudData measurement,
+    cartographer::mapping::TrajectoryBuilderInterface *trajectory_builder,
+    int trajectory_id, cartographer::transform::Rigid3d tmp_global_pose) {
```
EshaMaharishi, line 653: Per offline discussion:

1. We may need to take in the measurement as a different, C-compatible type in the future, but since we don't know yet what that type will be, we'll leave it as `cartographer::sensor::TimedPointCloudData`

2. We should make the `cartographer::mapping::TrajectoryBuilderInterface *trajectory_builder` and `int trajectory_id` members of SlamServiceImpl, so that they don't need to be passed in here

3. We should refactor the calling code so that `tmp_global_pose` doesn't need to be passed in here

EshaMaharishi (commented):
Looking good! Noted the conclusions from our discussion

viam-cartographer/src/slam_service/slam_service.h
```diff
@@ -252,6 +256,9 @@ class SLAMServiceImpl final : public SLAMService::Service {
     // started.
     std::string latest_pointcloud_map;
     // ---
+
+    cartographer::mapping::TrajectoryBuilderInterface *trajectory_builder;
+    int trajectory_id;
```
EshaMaharishi, line 261: @kkufieta , do we need to access `trajectory_builder` and `trajectory_id` under `map_builder_mutex`?

If so @kim-mishra could we move this to just after L244, so that it's grouped with things that need to be accessed under `map_builder_mutex`?

EshaMaharishi (approved):
LGTM mod comment!

viam-cartographer/src/slam_service/slam_service.h
kkufieta: Yes, `trajectory_builder` should be protected by `map_builder_mutex`. It makes sense to do the same for `trajectory_id` even though it's probably not strictly necessary.

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -647,10 +647,31 @@ void SLAMServiceImpl::SaveMapWithTimestamp() {
     }
 }
 
+void SLAMServiceImpl::AddSensorReading(
```
nicksanford, line 650: 1. This function should return a boolean (true for if the lock was acquired & the sensor data was written to cartographer & false if neither of those things happened).
2. This function should `attempt` to aquire the `map_builder_mutex` & give up if it is unable to acquire the lock.
3. I wasn't aware that the `viam_response_mutex` existed when I designed this api. I thought that there was only one mutex. I think we need to know what other code locks `viam_response_mutex` & how long this lock might block for. If it is always going to be nanoseconds I think this implementation is fine but if it could be longer, I think we need to discuss more what we do.

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -647,10 +647,31 @@ void SLAMServiceImpl::SaveMapWithTimestamp() {
...
+    cartographer::sensor::TimedPointCloudData measurement) {
+    cartographer::transform::Rigid3d tmp_global_pose;
+    bool update_latest_global_pose = false;
+
+    {
+        std::lock_guard<std::mutex> lk(map_builder_mutex);
+        if (measurement.ranges.size() > 0) {
```
nicksanford, line 657: Why do we need to check that the measurement ranges's size is greater than 0?

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -647,10 +647,31 @@ void SLAMServiceImpl::SaveMapWithTimestamp() {
...
+            trajectory_builder->AddSensorData(kRangeSensorId.id, measurement);
+            auto local_poses = map_builder.GetLocalSlamResultPoses();
+            if (local_poses.size() > 0) {
+                update_latest_global_pose = true;
+                tmp_global_pose = map_builder.GetGlobalPose(trajectory_id,
```
nicksanford, line 662: What is this trajectory_id? What sets it? Why do we need to provide it to get the global pose?

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -647,10 +647,31 @@ void SLAMServiceImpl::SaveMapWithTimestamp() {
...
+                                                            local_poses.back());
```
nicksanford, line 663: What does `local_poses.back()` return?

viam-cartographer/src/slam_service/slam_service.h
```diff
@@ -252,6 +256,9 @@ class SLAMServiceImpl final : public SLAMService::Service {
     // started.
     std::string latest_pointcloud_map;
     // ---
+
+    cartographer::mapping::TrajectoryBuilderInterface *trajectory_builder;
```
nicksanford, line 260: 1. Why does this need to be global?
2. Where is this pointer initialized?
3. Is it possible for `AddSensorReading` to be called before `ProcessDataAndStartSavingMaps` is called & causes memory corruption due to this pointer pointing at invalid memory?
4. When do we free the object this pointer points to?
5. If we really need a pointer on an object, could we construct this at CPU-bound initialization time (that way we are guaranteed it will always be a valid pointer) & have this be a smart pointer?

viam-cartographer/src/slam_service/slam_service.h
nicksanford: Can we please move this to CPU bound initialization time (i.e. point 5)?

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -741,6 +744,8 @@ void SLAMServiceImpl::ProcessDataAndStartSavingMaps(double data_start_time) {
     if (!use_live_data) {
         // We still want to optimize the map in localization mode, but we do not
         // need to update the backup of the map
+        cartographer::transform::Rigid3d tmp_global_pose;
```
nicksanford, line 747: 1. Why did we need to move this inside the loop?
2. Is there a risk of us leaking memory from creating new `tmp_global_pose`s every iteration of the while loop without ever freeing the one from the previous loop?

viam-cartographer/src/slam_service/slam_service.cc
kim-mishra: the last element of `local_poses`, its a [method on std::vector](https://cplusplus.com/reference/vector/vector/back/)

viam-cartographer/src/slam_service/slam_service.h
kim-mishra: 1. I believe we put this definition here so that it could be accessed within `AddSensorReading` so that we did not have to pass it in and change the function signature
2. I think it is initialized [here](https://github.com/viamrobotics/viam-cartographer/blob/93f5c18e56da635fb74a9d6f8896bb1815f7a9e0/viam-cartographer/src/slam_service/slam_service.cc#L657) within `map_builder.SetTrajectoryBuilder()`
3. I think that would be a problem yes, we talked about moving it into the initialization function @kkufieta is working on as a separate PR
4. I think that would happen in the terminate function [this ticket](https://viam.atlassian.net/browse/RSDK-3327) refers to
5. I am a little confused on what you mean here - would putting it in Kats init function be what you are talking aout?

viam-cartographer/src/slam_service/slam_service.cc
kim-mishra: 1. Because we wanted to keep the function signature simple and not have to pass this var into `AddSensorReading` we wanted to only initialize it if it were going to be used, although now I am getting a little confused why it is better for this to be in the loop. @kkufieta would you mind explaining this again
2. Based on this [doc](https://ptolemy.berkeley.edu/ptolemyclassic/almagest/docs/prog/html/ptlang.doc7.html) I think we actually will need to delete this if we continue with this route

viam-cartographer/src/slam_service/slam_service.cc
kim-mishra: I am looking into this but having a tough time figuring out why - it is leftover from the old code

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -647,10 +647,31 @@ void SLAMServiceImpl::SaveMapWithTimestamp() {
...
+    cartographer::sensor::TimedPointCloudData measurement) {
+    cartographer::transform::Rigid3d tmp_global_pose;
+    bool update_latest_global_pose = false;
+
+    {
+        std::lock_guard<std::mutex> lk(map_builder_mutex);
+        if (measurement.ranges.size() > 0) {
+            trajectory_builder->AddSensorData(kRangeSensorId.id, measurement);
+            auto local_poses = map_builder.GetLocalSlamResultPoses();
+            if (local_poses.size() > 0) {
```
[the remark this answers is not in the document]
nicksanford: That seems like a great reason. Thanks!

viam-cartographer/src/slam_service/slam_service.cc
nicksanford: Thanks for looking into this @kim-mishra ! I still have some more questions @kkufieta I see you write the map builder code, so I'm wondering if you might be able to help answer some of them (no worries if not, I know this code was written a while ago):
1. Why is it safe to only use a single trajectory_id for the entire mapping / updating / localizing session?
2. Under what conditions will `local_poses.size()` be 0?
3. What is `kRangeSensorId`?

viam-cartographer/src/slam_service/slam_service.cc
nicksanford: @kkufieta 
1. are we guaranteed that the most recent pose is always the last element of `local_poses`? 
2. Does this mean that `local_poses` grows unbounded over the lifetime of a given carto_grpc_server process?

viam-cartographer/src/slam_service/slam_service.cc
nicksanford: I think GetPointCloudMap will paint the entire map, which I think is the most expensive operation possible, isn't it. 

@kkufieta is there a reason why we use a single lock around both updating the map (which is big & slow) and the position (which is small & fast)?

It seems like we could likely gain a lot of performance by having different locks around these two different entities but please let me know if I'm missing something.

viam-cartographer/src/slam_service/slam_service.cc
kim-mishra: From what I can tell, the most recent pose is added to the back of `local_pose` and nothing is ever removed from it so yes and yes, but please correct me if I am wrong!

viam-cartographer/src/slam_service/slam_service.cc
kkufieta: I suspect this is here because, technically, a PCD could contain 0 points, which I suspect cartographer would complain about. This ensures that we only attempt to add measurements that contain data.

viam-cartographer/src/slam_service/slam_service.cc
kkufieta: 1. In our case, we're starting a new trajectory each time we start up cartographer, and we only need to keep track of that one trajectory (together with the associated `trajectory_id`. 
2. It can happen in the beginning when starting up cartographer. We observed that just because we added a sensor reading, `local_poses` did not necessarily have an entry yet.
3. `kRangeSensorId` defines the sensor type: https://github.com/cartographer-project/cartographer/blob/ef00de2317dcf7895b09f18cc4d87f8b533a019b/cartographer/mapping/trajectory_builder_interface.h#L66-L73

viam-cartographer/src/slam_service/slam_service.cc
kkufieta: 1. Yes, I do also believe that the most recent pose is always the last element of `local_poses`.
2. I am not sure, but my guess is "yes" - in the same manner as the overall map (pbstream) is growing unbounded

viam-cartographer/src/slam_service/slam_service.cc
kkufieta: This is happening outside the loop. This code is only executed in offline mode after we've processed all the data.

viam-cartographer/src/slam_service/slam_service.cc
kkufieta: I believe using two different locks was being considered, but since we're only copying a string or `Rigid3d` object, we assumed those operations are fast enough to be done under one shared mutex - and the code would be easier to maintain that way.

nicksanford: Thanks @kkufieta! I've updated my question

PR #103, opened 2023-05-26
RSDK-3276/RSDK-3277: Update cartographer for data replay

dmhilly:
### Summary
This PR covers the following two tickets:

1. [RSDK-3276](https://viam.atlassian.net/browse/RSDK-3276): viam-cartographer should read from sensor even when use_live_data = false, and not wait 200 ms
2. [RSDK-3277](https://viam.atlassian.net/browse/RSDK-3277): Cartographer should keep polling for new data even when use_live_data=false

It also modifies an existing test for offline mode that was broken by this change.

### Manual Testing
I tested with a dataset of about ~1000 images and confirmed all of the images were downloaded (by looking at the data directory) and fed into cartographer (by printing this out - technically all but the last one are fed in, which is expected given my implementation). I confirmed that the time taken to build the map is similar to offline mode (a couple seconds) and that the map is built correctly.

[RSDK-3276]: https://viam.atlassian.net/browse/RSDK-3276?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RSDK-3277]: https://viam.atlassian.net/browse/RSDK-3277?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -526,31 +526,40 @@ void SLAMServiceImpl::RunSLAM() {
...
+        // 1. Cartographer needs at least two PCD files to work properly.
+        // 2. A .DS_Store file is frequently added to the data directory when
+        // a user opens the directory on osx.
+        // Expecting a minimum of 3 files solves both problems without having to
+        // loop over and count the number of actual data files in the data
+        // directory.
+        if (file_list_offline.size() > 2) {
+            // If the current file is the last in the file list, skip it, since
+            // it's possible it's still being written to.
+            if (current_file_offline != file_list_offline.size() - 1) {
```
dmhilly, line 545: We'll never read the last file now, if that's okay. I couldn't think of a good way around this.

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -526,31 +526,40 @@ void SLAMServiceImpl::RunSLAM() {
...
+                const auto to_return = file_list_offline[current_file_offline];
+                if (to_return != "") {
```
dmhilly, line 547: This might not be necessary anymore actually; the last file in the list was showing up as "", not sure if that's a C++ null termination thing or a shortcoming of how our file list is being constructed.

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -526,31 +526,40 @@ void SLAMServiceImpl::RunSLAM() {
...
+                    current_file_offline++;
+                    return to_return;
+                }
+            }
+        }
+
+        // Sleep for a short period of time if we didn't find any files. This is
+        // to prevent cartographer from using up too much CPU in this loop when
+        // there's no new data coming in.
+        std::this_thread::sleep_for(data_rate_ms);
```
dmhilly, line 557: This is not strictly necessary but it seemed nice. We could add this to the data process loop too in `viam-cartographer.go`.

dmhilly:
@EshaMaharishi you can hold off on this review for now as I realize there are still C++ tests that need to be fixed. My apologies!

EshaMaharishi:
@dmhilly no worries, sounds good!

viam-cartographer/src/slam_service/slam_service.cc
tessavitabile: Will data_rate_ms be zero for actual offline mode? If not, this will cause actual offline to run unnecessarily slowly.

Also, I don't understand why we would add this to the data process loop in viam-cartographer.go, since that loop doesn't have to wait for new data to come in.

dmhilly: It shouldn't cause this to run slow because we only sleep if we were unable to find more files. For offline mode, this should only happen when we reach the end of the dataset. For the replay camera, this could happen if there replay camera is unable to keep up with cartographer and files are not being written fast enough. I added this because after this change, we never stop SLAM in offline mode/replay camera mode, so this should make this loop less resource-intensive once we've reached the end of the dataset.

viam-cartographer/src/slam_service/slam_service_test.cc
```diff
@@ -318,32 +318,6 @@ BOOST_AUTO_TEST_CASE(SetActionMode_invalid_case) {
     utils::removeTmpDirectory(tmp_dir);
 }
 
-BOOST_AUTO_TEST_CASE(GetNextDataFileOffline_not_enough_data) {
```
dmhilly, line 321: I ended up removing this test. We still will not try to construct a map with just two files, but instead we'll just keep looping until we find more files. I couldn't figure out how to test this since `GetNextDataFileOffline()` will just block.

viam-cartographer/src/slam_service/slam_service_test.cc
```diff
@@ -360,7 +334,7 @@ BOOST_AUTO_TEST_CASE(GetNextDataFileOffline) {
         utils::createTmpDirectoryAndAddFiles(data_files, map_files);
     slamService.path_to_data = tmp_dir.string() + "/data";
 
-    for (int i = 0; i < data_files.size(); i++) {
+    for (int i = 0; i < data_files.size() - 1; i++) {
```
dmhilly, line 337: We now ignore the last file in offline mode because it's possible that it's still being written to.

dmhilly (commented):
This is now RFAL @EshaMaharishi

viam-cartographer/src/slam_service/slam_service.cc
tessavitabile: Ah, got it. I think it's a little confusing to use data_rate_ms here, since how long we should wait doesn't have to do with the data rate. But also we'll stop writing to files in full modularization, so we don't need to think too carefully about what's the right thing to do here.

viam-cartographer/src/slam_service/slam_service.cc
EshaMaharishi: That makes sense, since we can't know if the last file has been finished being written to until we see a newer file.

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -526,31 +526,38 @@ void SLAMServiceImpl::RunSLAM() {
...
+        // reasons:
+        // 1. Cartographer needs at least two PCD files to work properly.
+        // 2. A .DS_Store file is frequently added to the data directory when
+        // a user opens the directory on osx.
+        // Expecting a minimum of 3 files solves both problems without having to
+        // loop over and count the number of actual data files in the data
+        // directory.
+        if (file_list_offline.size() > 2) {
+            // If the current file is the last in the file list, we skip it,
+            // since it's possible it's still being written to.
```
EshaMaharishi, line 544: [nit] Rather than saying "we skip it", can we say "we don't return it yet"?

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -526,31 +526,38 @@ void SLAMServiceImpl::RunSLAM() {
...
+
+        // Sleep for a short period of time if we didn't find any files. This is
+        // to prevent cartographer from using up too much CPU in this loop when
+        // there's no new data coming in.
+        std::this_thread::sleep_for(data_rate_ms);
+
         // This log line is needed by rdk integration tests.
-        LOG(INFO) << "Finished processing offline data";
-        return "";
+        VLOG(1) << "No more data available";
```
EshaMaharishi, line 558: [nit] Could we say "No new data found"?

viam-cartographer/src/slam_service/slam_service_test.cc
EshaMaharishi: Could we make the test check that if there are less than 2 files, "No new data found" is logged?

EshaMaharishi (commented):
Looks good! Only minor comments

dmhilly (commented):
Thanks for the feedback all, I think I've addressed everything. This is RFAL!

viam-cartographer.go
```diff
@@ -273,32 +274,43 @@ func (cartoSvc *cartographerService) StartDataProcess(
...
-					}
+					cartoSvc.getNextDataPoint(cancelCtx, lidar, c)
 				})
 			}
 		}
 	})
 }
 
+func (cartoSvc *cartographerService) getNextDataPoint(ctx context.Context, lidar lidar.Lidar, c chan int) {
+	if _, err := dim2d.GetAndSaveData(ctx, cartoSvc.dataDirectory, lidar, cartoSvc.logger); err != nil {
```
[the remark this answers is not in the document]
tessavitabile: Ok cool, so this will be handled as part of full modularization.

tessavitabile (approved):

viam-cartographer.go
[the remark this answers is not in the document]
dmhilly: Ah yeah I realize I didn't answer the question - afaik it will be handled either by Full Mod _or_ Cloud Story. cc @EshaMaharishi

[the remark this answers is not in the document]
EshaMaharishi: @tessavitabile yes, most likely it will be handled as part of cloud story, since that's where we'll bring back offline mode and start using the replay camera in earnest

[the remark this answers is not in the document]
tessavitabile: Thank you!

EshaMaharishi (approved):

EshaMaharishi (approved):
LGTM mod comment

PR #106, opened 2023-05-30
[RSDK-3002] Add workflow to bump dependencies

JohnN193:
new workflow to update dependencies of viam-cartographer once a week. Currently requires the "safe to test" label to be added manually. The workflow will skip if no updates are needed, as a quick way to prevent the rest of the workflow from running. Future PR will add a comment when the PR is generated. 

[Tested](https://github.com/viamrobotics/viam-cartographer/actions/workflows/bump_dependencies.yml) by setting the workflow to trigger on push. Heavily inspired by [typescript sdk](https://github.com/viamrobotics/viam-typescript-sdk/blob/workflow/update-protos/.github/workflows/update_protos.yml) and [homebrew-brews](https://github.com/viamrobotics/homebrew-brews/blob/main/.github/workflows/bump-versions.yml)

example PR: https://github.com/viamrobotics/viam-cartographer/pull/111 

ticket: https://viam.atlassian.net/browse/RSDK-3002


.github/workflows/bump_dependencies.yml
```diff
@@ -0,0 +1,53 @@
...
+        git init
+        git add .
+        sudo chown -R testbot .
+        sudo -u testbot bash -lc 'go get -u .'
+        GEN_DIFF=$(git status -s)
+
+        if [ -n "$GEN_DIFF" ]; then
+            echo 
+            git status
+
```
Otterverse, line 39:
```suggestion
```

.github/workflows/bump_dependencies.yml
```diff
@@ -0,0 +1,53 @@
...
+    container:
+      image: ghcr.io/viamrobotics/canon:amd64
+    timeout-minutes: 10
+
+    steps:
+    - name: Check out code
+      uses: actions/checkout@v3
+      with:
+        ref: ${{ github.head_ref }}
+        token: ${{ secrets.GIT_ACCESS_TOKEN }}
```
Otterverse, line 26:
```suggestion
```


I don't think this is required. You only need a ref when checking out additional code or branches that aren't what triggered the workflow in the first place. Likewise, the token shouldn't be needed since the default token can always read the repo for checkout.

.github/workflows/bump_dependencies.yml
```diff
@@ -0,0 +1,53 @@
...
+
+    - name: Update go dependencies
+      run: |
+        git init
+        git add .
```
Otterverse, line 31:
```suggestion
```
You don't need to init a new repo here. The one you checked out above should be fine.

.github/workflows/bump_dependencies.yml
```diff
@@ -0,0 +1,53 @@
...
+        sudo -u testbot bash -lc 'go get -u .'
+        GEN_DIFF=$(git status -s)
+
+        if [ -n "$GEN_DIFF" ]; then
+            echo 
+            git status
+            
+        else
+            echo "No dependency updates needed"
+            exit 1
```
Otterverse, line 42: This is going to fail a run. You shouldn't fail just because no update happens. You want to just NOT do the next part. You should output a value in the first part of this if statement, (no need for the "else") then use an "if" in the following step to only do the PR if the value was set.

So something like `echo needs_pr=1 >> $GITHUB_OUTPUT` here, and then in the next step, `if: steps.goupdate.output.needs_pr == 1` (My syntax may not be perfect here.) Also, you'll need to add an `id: goupdate` field to this step (or whatever you want to id the step as.)

https://rmannibucau.metawerx.net/github-action-output-communicate-between-steps.html has a decent example.

.github/workflows/bump_dependencies.yml
```diff
@@ -0,0 +1,53 @@
...
+
+    - name: Add + Commit + Open PR
+      uses: peter-evans/create-pull-request@v5
+      with:
+          commit-message: '[WORKFLOW] Updating go dependencies'
+          branch: 'workflow/update-dependencies'
+          delete-branch: true
+          base: main
+          title: Automated Go Dependencies Update
+          body: This is an auto-generated PR to update go dependencies. Please confirm tests are passing before merging
```
Otterverse, line 53:
```suggestion
          body: This is an auto-generated PR to update go dependencies. Please confirm tests are passing before merging.
```

Otterverse (changes requested):
A few small changes suggested. Some just to simplify stuff, but the big one is you need to not have this just fail when there are no changes.

EshaMaharishi (commented):
Generally looks good to me! Every other Wednesdsay sounds good, the 10 minute timeout makes sense, and glad we're handling the "no update was needed" case. Moving myself to assignee since I don't have any comments. Thank you @JohnN193 and @Otterverse !

Otterverse (approved):
LGTM!

PR #121, opened 2023-06-05
RSDK-3346 - getInternalState refactor

kim-mishra:


viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -146,30 +154,25 @@ ::grpc::Status SLAMServiceImpl::GetPointCloudMap(
...
-        bool ok = map_builder.SaveMapToFile(true, filename);
-        if (!ok) {
-            std::ostringstream oss;
-            oss << "Failed to save the state as a pbstream.";
-            return grpc::Status(grpc::StatusCode::UNAVAILABLE, oss.str());
-        }
+    viam_carto_get_internal_state_response vcgicr;
+    int status = GetInternalStateC(&vcgicr);
+
+    if (status == 1) {
```
nicksanford, line 160: When this method is added to CartoFacade, we should be using the error code constants defined here rather than raw integers:
https://github.com/viamrobotics/viam-cartographer/blob/main/viam-cartographer/src/carto_facade/carto_facade.h#L87

viam-cartographer/src/slam_service/slam_service.cc
```diff
@@ -186,6 +197,36 @@ ::grpc::Status SLAMServiceImpl::GetInternalState(
...
+    }
+
+    std::string buf;
+    // deferring reading the pbstream file in chunks until we run into issues
+    // with loading the file into memory
+    try {
+        ConvertSavedMapToStream(filename, &buf);
+    } catch (std::runtime_error &e) {
+        LOG(ERROR) << e.what();
+        return VIAM_CARTO_SERIALIZATION_ERROR;
```
kim-mishra, line 220: If we don't want to catch all `runtime_errors`, we can change the type of exception thrown just thought we would want to change as little as possible for now, but let me know if you think we should be throwing more specific errors

nicksanford: If we needed to handle different runtime errors with different return codes we could look into creating a class per runtime error, however what you have here is appropriate for our current use case.

nicksanford (approved):
This looks good to me. Holding off on approval until this is moved to be on CartoFacade.

nicksanford (approved):
This looks good to me. Holding off on approval until this is moved to be on CartoFacade.

nicksanford (changes requested):
This looks good to me. Holding off on approval until this is moved to be on CartoFacade.

PR #123, opened 2023-06-05
RSDK-3362: Add test showing cartographer module looks for timestamps in context …

dmhilly:
…metadata

### DIFF
--- a/internal/testhelper/testhelper.go
+++ b/internal/testhelper/testhelper.go
@@ -20,6 +20,7 @@ import (
 	"go.viam.com/rdk/rimage/transform"
 	"go.viam.com/rdk/services/slam"
 	"go.viam.com/rdk/testutils/inject"
+	"go.viam.com/rdk/utils/contextutils"
 	"go.viam.com/test"
 	"go.viam.com/utils/artifact"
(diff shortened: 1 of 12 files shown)


internal/testhelper/testhelper.go
```diff
@@ -81,9 +85,15 @@ func SetupDeps(sensors []string) resource.Dependencies {
 	return deps
 }
 
-func getGoodLidar() *inject.Camera {
+func getValidSensor(isReplay bool) *inject.Camera {
```
EshaMaharishi, line 88: [opt] I think it'd be better to follow the pattern of a separate helper (`getReplaySensor`) than to pass a bool in, especially since the duplicated logic would be very straightforward and is also duplicated in the other helpers

EshaMaharishi (approved):
Lgtm with suggestion. Thanks for adding this test!

internal/testhelper/testhelper.go
dmhilly: SGTM, made that change!

EshaMaharishi: ty!

PR #125, opened 2023-06-05
RSDK-3425-CPU-bound-init-term

nicksanford:
[Ticket](https://viam.atlassian.net/browse/RSDK-3425)


viam-cartographer/src/carto_facade/carto_facade.cc
```diff
@@ -25,6 +49,21 @@ config from_viam_carto_config(viam_carto_config vcc) {
     c.map_rate_sec = vcc.map_rate_sec;
     c.mode = vcc.mode;
     c.lidar_config = vcc.lidar_config;
+    if (c.sensors.size() == 0) {
+        throw VIAM_CARTO_SENSORS_LIST_EMPTY;
+    }
+    if (c.data_dir.size() == 0) {
+        throw VIAM_CARTO_DATA_DIR_EMPTY;
+    }
```
kim-mishra, line 57: Should this be called something like `VIAM_CARTO_DATA_DIR_NOT_PROVIDED`?  `VIAM_CARTO_DATA_DIR_EMPTY` makes me think that the directory provided has nothing in it which would not be an issue right?

kim-mishra (approved):
LGTM have a small nit

viam-cartographer/src/carto_facade/carto_facade.cc
nicksanford: Great suggestion! Done.

PR #127, opened 2023-06-06
RSDK-3426-io-bound-init-term

nicksanford:
[Ticket](https://viam.atlassian.net/browse/RSDK-3426)

- Add CartoFacade.IOInit instance method which does the following:

1. detects if the data_dir has a deprecated format & throws if it does
2. creates the data_dir with the correct format if it doesn't exist
3. sets the correct action mode
4. creates & configures the map builder with the right hyperparameters based on the action mode (default hyperparameter config will live in go so that is not handled here)
5. starts the trajectory builder

- Add tests that verify IOInit setup (trying to keep the same level of coverage covered by viam::config & viam::SlamService)
- Call CartoFacade.IOInit within viam_carto_init C API method & catch any errors that IOInit throws, returning them as status codes
- Change Full Mod C++ tests to compile with debug information (for easier stacktrace debugging)
- Remove viam-cartographer/scripts/test_cartographer.sh as it is no longer needed
- Remove `viam_carto_MODE` as I realized that the mode will be inferred from map_rate_sec & the state of the file system
- add viam::carto_facade::io namespace functions (copied over from file_handler.cc) with tests which only run during Full Mod test cases
- add viam::carto_facade::MapBuilder, based on the old map builder which properly handles destruction & which now owns the pointer to trajectory builder
- add viam::carto_facade::util (copied from viam::utils)

PR #136, opened 2023-06-12
RSDK-3604-install-artifact

nicksanford:
[Ticket](https://viam.atlassian.net/browse/RSDK-3604)
This is work required to unblock: https://github.com/viamrobotics/viam-cartographer/pull/127/files

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,14 @@
 BUILD_CHANNEL?=local
 TOOL_BIN = bin/gotools/$(shell uname -s)-$(shell uname -m)
-PATH_WITH_TOOLS="`pwd`/$(TOOL_BIN):${PATH}"
+PATH_WITH_TOOLS="`pwd`/$(TOOL_BIN):$(HOME)/go/bin/:${PATH}"
 GIT_REVISION = $(shell git rev-parse HEAD | tr -d '\n')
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -71,6 +76,7 @@ ifeq ("Darwin", "$(shell uname -s)")
 else
 	cd viam-cartographer/scripts && ./setup_cartographer_linux.sh
 endif
+	@make artifact-pull
```
jeremyrhyde, line 79: [opt] It might be better to only run `make artifact-pull ` during `make test` (line 111). Potentially could also split off a new `make test-setup` if desired

jeremyrhyde (commented):
LGTM, I made one suggestion I'd like to come to a conclusion one before approving but should be the only outstanding point

Makefile
nicksanford: We don't want to run it every time we run tests, otherwise it will slow down tests (which are run frequently during development) Rather than adding new commands the entire slam team needs to remember to run whenever they reinstall the system I decided to slightly slow down `make setup` (which should be run infrequently on local machines) in order to ensure that tests would "just work" for the entire team after this change & not require the entire team to remember to run a new make command.

kim-mishra (approved):

jeremyrhyde (approved):
Question resovled! LGTM

PR #139, opened 2023-06-14
RSDK-3143 - CGO api implementation for full mod

kim-mishra:
[Ticket](https://viam.atlassian.net/browse/RSDK-3143)
- adds API to call into cartofacade
- adds tests for API
- only includes C++ library imports if C++ compiler is being used

also manually tested that the data passed in C are what we send


cartofacade/cgo_api.go
```diff
@@ -0,0 +1,400 @@
...
+	#include "../viam-cartographer/src/carto_facade/carto_facade.h"
+
+	viam_carto_lib* alloc_viam_carto_lib() { return (viam_carto_lib*) malloc(sizeof(viam_carto_lib)); }
+	void free_alloc_viam_carto_lib(viam_carto_lib* p) { free(p); }
+
+	viam_carto* alloc_viam_carto() { return (viam_carto*) malloc(sizeof(viam_carto)); }
+	void free_alloc_viam_carto(viam_carto* p) { free(p); }
+
+	bstring* alloc_bstring_array(size_t len) { return (bstring*) malloc(len * sizeof(bstring)); }
+	void free_bstring_array(bstring* p) { free(p); }
```
kim-mishra, line 17: @nicksanford I just remembered you mentioned that we might need to handle a case where C could not allocate memory here. Do you have any suggestions on how best to do it?

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,400 @@
+// Package cartofacade provides an api to call into c code
+package cartofacade
```
nicksanford, line 2:
```suggestion
//nolint:lll
// Package cartofacade provides an api to call into c code
package cartofacade
```
Add this to disable line length linting

nicksanford (changes requested):

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,400 @@
...
..._random_distributions  -labsl_random_seed_sequences  -labsl_random_internal_pool_urbg  -labsl_random_internal_randen  -labsl_random_internal_randen_hwaes  -labsl_random_internal_randen_hwaes_impl  -labsl_random_internal_randen_slow  -labsl_random_internal_platform  -labsl_random_internal_seed_material  -labsl_random_seed_gen_exception  -labsl_statusor  -labsl_status  -labsl_cord  -labsl_cordz_info  -labsl_cord_internal  -labsl_cordz_functions  -labsl_exponential_biased  -labsl_cordz_handle  -labsl_crc_cord_state  -labsl_crc32c  -labsl_crc_internal  -labsl_crc_cpu_detect  -labsl_bad_optional_access  -labsl_strerror  -labsl_str_format_internal  -labsl_synchronization  -labsl_graphcycles_internal  -labsl_stacktrace  -labsl_symbolize  -labsl_debugging_internal  -labsl_demangle_internal  -labsl_malloc_internal  -labsl_time  -labsl_civil_time  -labsl_strings  -labsl_strings_internal  -labsl_base  -labsl_spinlock_wait  -labsl_int128  -labsl_throw_delegate  -labsl_time_zone  -labsl_bad_variant_access  -labsl_raw_logging_internal  -labsl_log_severity -lpthread  -lgrpc++ -lcairo -llua -lstdc++ -lceres -lprotobuf -lglog -lboost_system -lboost_filesystem -lboost_iostreams -lpcl_io -lpcl_common
```
nicksanford, line 6: Are you certain that you need all of these packages?

cartofacade/cgo_api.go
nicksanford: Yes. if `alloc_viam_carto` returns `nullptr` then it means that allocation has failed & the go function which called it should return early with an error.

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,401 @@
...
+	value *C.viam_carto
+}
+
+// AddSensorReading holds the c type viam_carto_sensor_reading
+type AddSensorReading struct {
+	value C.viam_carto_sensor_reading
+}
+
+// GetPositionResponse holds the c type viam_carto_get_position_response
+type GetPositionResponse struct {
```
nicksanford, line 45:
```suggestion
type getPositionResponse struct {
```

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,401 @@
...
+	value *C.viam_carto_lib
+}
+
+// Carto holds the c type viam_carto
+type Carto struct {
+	value *C.viam_carto
+}
+
+// AddSensorReading holds the c type viam_carto_sensor_reading
+type AddSensorReading struct {
```
nicksanford, line 40: Can we make any types that are internal to the package private?
```suggestion
type addSensorReading struct {
```

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,401 @@
...
+	value C.viam_carto_sensor_reading
+}
+
+// GetPositionResponse holds the c type viam_carto_get_position_response
+type GetPositionResponse struct {
+	value C.viam_carto_get_position_response
+}
+
+// GetPositionHolder holds values returned from c to be processed later
+type GetPositionHolder struct {
```
nicksanford, line 50:
```suggestion
type GetPosition struct {
```

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,401 @@
...
+	real float64
+	imag float64
+	jmag float64
+	kmag float64
+
+	compReference string
+}
+
+// GetInternalStateResponse holds the c type viam_carto_get_internal_state_response
+type GetInternalStateResponse struct {
```
nicksanford, line 69:
```suggestion
type getInternalStateResponse struct {
```

cartofacade/cgo_api.go
nicksanford: Also could we please organize the private types under the public ones?

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,401 @@
...
+	oy    float64
+	oz    float64
+	theta float64
+
+	real float64
+	imag float64
+	jmag float64
+	kmag float64
+
+	compReference string
```
nicksanford, line 65:
```suggestion
	componentReference string
```
Also, are you sure these fields don't need to be public so the caller can read them?

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,401 @@
...
+type LidarConfig int64
+
+const (
+	twoD LidarConfig = iota
+	threeD
+)
+
+// CartoConfig contains config values from app
+type CartoConfig struct {
+	sensors            []string
```
nicksanford, line 83: Same question regarding whether these fields need to be public for the caller to set them

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,401 @@
...
+	minAddedSubmapsCount int
+	occupiedSpaceWeight  float64
+	translationWeight    float64
+	rotationWeight       float64
+}
+
+// NewCartoLib calls viam_carto_lib_init and returns a pointer to a viam carto lib object.
+func NewCartoLib(miniloglevel, verbose int) (CartoLib, error) {
+	pVcl := C.alloc_viam_carto_lib()
+	defer C.free_alloc_viam_carto_lib(pVcl)
```
nicksanford, line 110: We don't want to defer this if the allocation fails

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,401 @@
...
+					// ppVcl := (**C.viam_carto_lib)(goPtr)
+					fmt.Println(ppVcl) = 0x600001ee8000
+	*/
+
+	status := C.viam_carto_lib_init(ppVcl, C.int(miniloglevel), C.int(verbose))
+	if err := getErrorFromStatusCode(status); err != nil {
+		return CartoLib{}, err
+	}
+
+	vcl := CartoLib{value: (*ppVcl)}
```
nicksanford, line 127: Why are we storing what`ppVcl` points to, rather than `ppVcl` itself?

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,401 @@
...
+
+	return vcl, nil
+}
+
+// TerminateCartoLib calls viam_carto_lib_terminate to clean up memory for viam carto lib.
+func (vcl *CartoLib) TerminateCartoLib() error {
+	status := C.viam_carto_lib_terminate((**C.viam_carto_lib)(unsafe.Pointer(vcl)))
```
nicksanford, line 134: I don't understand how this can work. `vcl` is a `CartoLib{}` which is not a valid input into `viam_carto_lib_terminate`.

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,401 @@
...
+	if err := getErrorFromStatusCode(status); err != nil {
+		return err
+	}
+	return nil
+}
+
+// NewCarto calls viam_carto_init and returns a pointer to a viam carto object.
+func NewCarto(cfg CartoConfig, acfg CartoAlgoConfig, vcl CartoLib) (Carto, error) {
+	pVc := C.alloc_viam_carto()
```
nicksanford, line 143: we need to early return if this allocation fails & not defer the free

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,401 @@
...
+	case twoD:
+		return C.VIAM_CARTO_TWO_D
+	case threeD:
+		return C.VIAM_CARTO_THREE_D
+	}
+	// Question: how should we represent an error case?
+	return 0
+}
+
+func getConfig(cfg CartoConfig) C.viam_carto_config {
```
nicksanford, line 322:
```suggestion
func toConfig(cfg CartoConfig) C.viam_carto_config {
```

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,401 @@
...
+		goStrings = append(goStrings, C.GoString(C.bstr2cstr(bString, 0)))
+	}
+	return goStrings
+}
+
+func bstringToGoString(bstr C.bstring) string {
+	return C.GoString(C.bstr2cstr(bstr, 0))
+}
+
+func getLidarConfig(lidarConfig LidarConfig) C.viam_carto_LIDAR_CONFIG {
```
nicksanford, line 311:
```suggestion
func toLidarConfig(lidarConfig LidarConfig) C.viam_carto_LIDAR_CONFIG {
```

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,401 @@
...
+	vcc.sensors_len = C.int(sz)
+	vcc.map_rate_sec = C.int(cfg.mapRateSecond)
+	vcc.data_dir = C.bfromcstr(C.CString(cfg.dataDir))
+	vcc.component_reference = C.bfromcstr(C.CString(cfg.componentReference))
+	vcc.lidar_config = getLidarConfig(cfg.lidarConfig)
+
+	return vcc
+}
+
+func getAlgoConfig(acfg CartoAlgoConfig) C.viam_carto_algo_config {
```
nicksanford, line 343:
```suggestion
func toAlgoConfig(acfg CartoAlgoConfig) C.viam_carto_algo_config {
```

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,401 @@
...
+	vcac.fresh_submaps_count = C.int(acfg.freshSubmapsCount)
+	vcac.min_covered_area = C.double(acfg.minCoveredArea)
+	vcac.min_added_submaps_count = C.int(acfg.minAddedSubmapsCount)
+	vcac.occupied_space_weight = C.double(acfg.occupiedSpaceWeight)
+	vcac.translation_weight = C.double(acfg.translationWeight)
+	vcac.rotation_weight = C.double(acfg.rotationWeight)
+	return vcac
+}
+
+// FreeBstringArray used to cleanup a bstring array (needs to be a go func so it can be used in tests)
```
nicksanford, line 361:
```suggestion
// FreeBstringArray used to cleanup a bstring array (needs to be a public function so it can be used in tests)
```

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,401 @@
...
+func FreeBstringArray(arr *C.bstring) {
+	C.free_bstring_array(arr)
+}
+
+func getErrorFromStatusCode(status C.int) error {
```
nicksanford, line 366:
```suggestion
func toError(status C.int) error {
```

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,401 @@
...
+	switch int(status) {
+	case C.VIAM_CARTO_SUCCESS:
+		return nil
+	case C.VIAM_CARTO_UNABLE_TO_ACQUIRE_LOCK:
+		return errors.New("error: VIAM_CARTO_UNABLE_TO_ACQUIRE_LOCK")
```
nicksanford, line 371: I don't think you need the `error: ` prefix

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,401 @@
...
+	defer C.free_bstring_array(vcc.sensors)
+
+	vcac := getAlgoConfig(acfg)
+
+	status := C.viam_carto_init(ppVc, vcl.value, vcc, vcac)
+	if err := getErrorFromStatusCode(status); err != nil {
+		return Carto{}, err
+	}
+
+	return Carto{value: (*ppVc)}, nil
```
nicksanford, line 158: Why are we storing whatppVcl points to, rather than ppVc itself?

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,401 @@
...
+	if err := getErrorFromStatusCode(status); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// TerminateCarto calls viam_carto_terminate to clean up memory for viam carto
+func (vc *Carto) TerminateCarto(ctx context.Context) error {
+	status := C.viam_carto_terminate(&vc.value)
```
nicksanford, line 185: Why does this not require a cast to unsafe pointer? If this works, why are we not using this approach in the lib terminate?

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,401 @@
...
+		return CartoLib{}, err
+	}
+
+	vcl := CartoLib{value: (*ppVcl)}
+
+	return vcl, nil
+}
+
+// TerminateCartoLib calls viam_carto_lib_terminate to clean up memory for viam carto lib.
+func (vcl *CartoLib) TerminateCartoLib() error {
```
nicksanford, line 133:
```suggestion
func (vcl *CartoLib) Terminate() error {
```

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,401 @@
...
+	status := C.viam_carto_lib_terminate((**C.viam_carto_lib)(unsafe.Pointer(vcl)))
+	if err := getErrorFromStatusCode(status); err != nil {
+		return err
+	}
+	return nil
+}
+
+// NewCarto calls viam_carto_init and returns a pointer to a viam carto object.
+func NewCarto(cfg CartoConfig, acfg CartoAlgoConfig, vcl CartoLib) (Carto, error) {
```
nicksanford, line 142:
```suggestion
func New(cfg CartoConfig, acfg CartoAlgoConfig, vcl CartoLib) (Carto, error) {
```

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,401 @@
...
+
+	if err := getErrorFromStatusCode(status); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// TerminateCarto calls viam_carto_terminate to clean up memory for viam carto
+func (vc *Carto) TerminateCarto(ctx context.Context) error {
```
nicksanford, line 184:
```suggestion
func (vc *Carto) Terminate(ctx context.Context) error {
```

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,401 @@
...
+		return err
+	}
+
+	return nil
+}
+
+// AddSensorReading is a wrapper for viam_carto_add_sensor_reading
+func (vc *Carto) AddSensorReading(ctx context.Context, readings []byte, timestamp time.Time) error {
+	sensorReading := C.viam_carto_sensor_reading{}
+	sensorReading.sensor_reading = C.blk2bstr(unsafe.Pointer(&readings[0]), C.int(len(readings)))
```
nicksanford, line 197: We should get confirmation from Eric that this is the right way to do this

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,401 @@
...
+
+	resp := AddSensorReading{value: sensorReading}
+
+	status := C.viam_carto_add_sensor_reading(vc.value, &resp.value)
+
+	if err := getErrorFromStatusCode(status); err != nil {
+		return err
+	}
+
+	status = C.viam_carto_add_sensor_reading_destroy(&resp.value)
```
nicksanford, line 209: This needs to be deferred after `viam_carto_add_sensor_reading` so that we do it even if we get an error from adding a sensor reading

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,401 @@
...
+
+		real: float64(resp.value.real),
+		imag: float64(resp.value.imag),
+		jmag: float64(resp.value.jmag),
+		kmag: float64(resp.value.kmag),
+
+		compReference: bstringToGoString(resp.value.component_reference),
+	}
+
+	status = C.viam_carto_get_position_response_destroy(&resp.value)
```
nicksanford, line 245: needs to be deferred

nicksanford (changes requested):

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,401 @@
...
+	if err := getErrorFromStatusCode(status); err != nil {
+		return nil, err
+	}
+
+	bytes := (*[]byte)(unsafe.Pointer(&resp.value.internal_state))
+	return *bytes, nil
+}
+
+func bStringToGoStringSlice(bstr *C.bstring, length int) []string {
+	bStrings := (*[1 << 28]C.bstring)(unsafe.Pointer(bstr))[:length:length]
```
nicksanford, line 298: Its not obvious to my why this works & whether or not it is correct. Could you please explain?

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,401 @@
...
+	status = C.viam_carto_get_internal_state_response_destroy(&resp.value)
+	if err := getErrorFromStatusCode(status); err != nil {
+		return nil, err
+	}
+
+	bytes := (*[]byte)(unsafe.Pointer(&resp.value.internal_state))
+	return *bytes, nil
+}
+
+func bStringToGoStringSlice(bstr *C.bstring, length int) []string {
```
nicksanford, line 297: If this is only used in testing, could we please put it in the test file?

nicksanford (changes requested):

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,401 @@
...
+}
+
+// AddSensorReading is a wrapper for viam_carto_add_sensor_reading
+func (vc *Carto) AddSensorReading(ctx context.Context, readings []byte, timestamp time.Time) error {
+	sensorReading := C.viam_carto_sensor_reading{}
+	sensorReading.sensor_reading = C.blk2bstr(unsafe.Pointer(&readings[0]), C.int(len(readings)))
+
+	sensorReading.sensor_reading_time_unix_micro = C.ulonglong(timestamp.UnixMicro())
+
+	resp := AddSensorReading{value: sensorReading}
```
nicksanford, line 213: Why are we wrapping `sensorReading` in a `AddSensorReading{}`?

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,401 @@
...
+	status = C.viam_carto_get_position_response_destroy(&resp.value)
+	if err := getErrorFromStatusCode(status); err != nil {
+		return GetPositionHolder{}, err
+	}
+
+	return getPositionHolder, nil
+}
+
+// GetPointCloudMapResponse is a struct to hold the c type viam_carto_get_point_cloud_map_response
+type GetPointCloudMapResponse struct {
```
nicksanford, line 266: types should be at the top.

cartofacade/cgo_api.go
kim-mishra: they should be public, updating

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,401 @@
...
+	if err := getErrorFromStatusCode(status); err != nil {
+		return nil, err
+	}
+
+	status = C.viam_carto_get_point_cloud_map_response_destroy(&resp.value)
+	if err := getErrorFromStatusCode(status); err != nil {
+		return nil, err
+	}
+
+	bytes := (*[]byte)(unsafe.Pointer(&resp.value.point_cloud_pcd))
```
nicksanford, line 286: I don' think this is correct... `point_cloud_pcd` is a `bstring`, shouldn't we be converting the better string into a go byte slice? Where did you find docs on how to do this?

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,401 @@
...
+	case C.VIAM_CARTO_SLAM_MODE_INVALID:
+		return errors.New("error: VIAM_CARTO_SLAM_MODE_INVALID ")
+	case C.VIAM_CARTO_LIDAR_CONFIG_INVALID:
+		return errors.New("error: VIAM_CARTO_LIDAR_CONFIG_INVALID ")
+	case C.VIAM_CARTO_MAP_RATE_SEC_INVALID:
+		return errors.New("error: VIAM_CARTO_MAP_RATE_SEC_INVALID ")
+	case C.VIAM_CARTO_COMPONENT_REFERENCE_INVALID:
+		return errors.New("error: VIAM_CARTO_COMPONENT_REFERENCE_INVALID ")
+	default:
+		return errors.New("error: unknown status code")
```
nicksanford, line 414: This should use language that is different from `VIAM_CARTO_UNKNOWN_ERROR`. This should be something like `status code unclassified`

cartofacade/cgo_api.go
kim-mishra: yes, updating

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,416 @@
...
+	status := C.viam_carto_init(ppVc, vcl.value, vcc, vcac)
+	if err := getErrorFromStatusCode(status); err != nil {
+		return Carto{}, err
+	}
+
+	return Carto{value: (*ppVc)}, nil
+}
+
+// Start is a wrapper for viam_carto_start
+func (vc *Carto) Start(ctx context.Context) error {
```
nicksanford, line 174: Why do all these functions take a context which they don't use? Can we please remove all context parameters to the CGO api?

nicksanford (changes requested):

cartofacade/cgo_api.go
kim-mishra: this was definitely a mistake, I updated this line to be:

`status := C.viam_carto_lib_terminate((**C.viam_carto_lib)(unsafe.Pointer(&vcl.value)))`

I suspect casting the untyped memory is why it worked, but it definitely would have been incorrect and think that we should ask eric about this as well

cartofacade/cgo_api.go
kim-mishra: This is the type that is needed for all operations aside from initialization so I thought it would be best to store this instead of getting what ppVcl points to each time we want to call any method. What do you think?

cartofacade/cgo_api.go
kim-mishra: updated to use this approach, the unsafe pointer was unnecessary I believe

cartofacade/cgo_api.go
kim-mishra: When I do that, I get an error that I cannot use cgo in testing files

cartofacade/cgo_api.go
nicksanford: I think terminate also takes a ppVcl

kim-mishra: right, so for only terminate I would get a reference to this obj instead of getting what this obj points to for each other method. Happy to change it either way to what you think is best

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,401 @@
...
+	resp := GetPositionResponse{value: C.viam_carto_get_position_response{}}
+
+	status := C.viam_carto_get_position(vc.value, &resp.value)
+
+	if err := getErrorFromStatusCode(status); err != nil {
+		return GetPositionHolder{}, err
+	}
+
+	getPositionHolder := GetPositionHolder{
+		x: float64(resp.value.x),
```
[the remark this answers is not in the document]
kim-mishra: they are all doubles in the C code, which are 8 bytes according to [here](https://www.tutorialspoint.com/cprogramming/c_data_types.htm)

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,401 @@
...
+}
+
+func getAlgoConfig(acfg CartoAlgoConfig) C.viam_carto_algo_config {
+	vcac := C.viam_carto_algo_config{}
+	vcac.optimize_on_start = C.bool(acfg.optimizeOnStart)
+	vcac.optimize_every_n_nodes = C.int(acfg.optimizeEveryNNodes)
+	vcac.num_range_data = C.int(acfg.numRangeData)
+	vcac.missing_data_ray_length = C.float(acfg.missingDataRayLength)
+	vcac.max_range = C.float(acfg.maxRange)
+	vcac.min_range = C.float(acfg.minRange)
```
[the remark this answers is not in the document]
kim-mishra: updated values that were floats in c to be to be float32s in Go, thanks for this catch!

cartofacade/cgo_api.go
kim-mishra: removed

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,456 @@
+// Package cartofacade provides an api to call into c code
+//
+//nolint:lll
+package cartofacade
+
+/*
+	#cgo CFLAGS: -I../viam-cartographer/src/carto_facade
+
+	// the libraries that need to be linked can be dermined by checking line 258 of the build.ninja file that is autogenerated during make build
+  	#cgo LDFLAGS: -L../viam-cartographer/build -L../viam-cartographer/cartographer/build -lcartographer  -lviam-cartographer  -labsl_raw_hash_set -labsl_hash  -labsl_str_format_internal  -labsl_synchronization    -labsl_strings  -labsl_throw_delegate -lcairo -llua -lstdc++ -lceres -lprotobuf -lglog -lboost_filesystem -lboost_iostreams -lpcl_io -lpcl_common
```
nicksanford, line 10: 👏 thank you very much for reducing this list down

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,456 @@
...
+	#include "../viam-cartographer/src/carto_facade/carto_facade.h"
+
+	viam_carto_lib* alloc_viam_carto_lib() { return (viam_carto_lib*) malloc(sizeof(viam_carto_lib)); }
+	void free_alloc_viam_carto_lib(viam_carto_lib* p) { free(p); }
+
+	viam_carto* alloc_viam_carto() { return (viam_carto*) malloc(sizeof(viam_carto)); }
+	void free_alloc_viam_carto(viam_carto* p) { free(p); }
+
+	bstring* alloc_bstring_array(size_t len) { return (bstring*) malloc(len * sizeof(bstring)); }
+	void free_bstring_array(bstring* p, int len) {
```
nicksanford, line 21: to make the  length type symmetrical with alloc
```suggestion
	void free_bstring_array(bstring* p, size_t len) {
```

nicksanford: Also, I'm very sorry to say that `bdestroy` is not a void function & in fact returns errors if called with invalid inputs.
Could you please propagate the error returned from bdestroy if it returns `BSTR_ERR` & `BSTR_OK` on success & check the error in go?

nicksanford (changes requested):

cartofacade/cgo_api.go
nicksanford: Nope, it doesn't need to change. I think this works. Thanks for clarifying

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,456 @@
...
+	value *C.viam_carto_lib
+}
+
+// Carto holds the c type viam_carto
+type Carto struct {
+	value *C.viam_carto
+}
+
+// getPointCloudMapResponse is a struct to hold the c type viam_carto_get_point_cloud_map_response
+type getPointCloudMapResponse struct {
```
nicksanford, line 47: Can we please move this to the private types?

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,456 @@
...
+/*
+	#cgo CFLAGS: -I../viam-cartographer/src/carto_facade
+
+	// the libraries that need to be linked can be dermined by checking line 258 of the build.ninja file that is autogenerated during make build
+  	#cgo LDFLAGS: -L../viam-cartographer/build -L../viam-cartographer/cartographer/build -lcartographer  -lviam-cartographer  -labsl_raw_hash_set -labsl_hash  -labsl_str_format_internal  -labsl_synchronization    -labsl_strings  -labsl_throw_delegate -lcairo -llua -lstdc++ -lceres -lprotobuf -lglog -lboost_filesystem -lboost_iostreams -lpcl_io -lpcl_common
+
+	#include "../viam-cartographer/src/carto_facade/carto_facade.h"
+
+	viam_carto_lib* alloc_viam_carto_lib() { return (viam_carto_lib*) malloc(sizeof(viam_carto_lib)); }
+	void free_alloc_viam_carto_lib(viam_carto_lib* p) { free(p); }
```
nicksanford, line 15:
```suggestion
	void free_viam_carto_lib(viam_carto_lib* p) { free(p); }
```

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,456 @@
...
+
+	viam_carto* alloc_viam_carto() { return (viam_carto*) malloc(sizeof(viam_carto)); }
+	void free_alloc_viam_carto(viam_carto* p) { free(p); }
```
nicksanford, line 18:
```suggestion
	void free_viam_carto(viam_carto* p) { free(p); }
```

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,456 @@
...
+	value C.viam_carto_get_internal_state_response
+}
+
+// NewLib calls viam_carto_lib_init and returns a pointer to a viam carto lib object.
+func NewLib(miniloglevel, verbose int) (CartoLib, error) {
+	pVcl := C.alloc_viam_carto_lib()
+	if pVcl == nil {
+		return CartoLib{}, errors.New("unable to allocate memory for viam carto lib")
+	}
+	defer C.free_alloc_viam_carto_lib(pVcl)
```
nicksanford, line 120:
```suggestion
	defer C.free_viam_carto_lib(pVcl)
```

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,456 @@
...
+	return nil
+}
+
+// New calls viam_carto_init and returns a pointer to a viam carto object.
+func New(cfg CartoConfig, acfg CartoAlgoConfig, vcl CartoLib) (Carto, error) {
+	pVc := C.alloc_viam_carto()
+	if pVc == nil {
+		return Carto{}, errors.New("unable to allocate memory for viam carto")
+	}
+	defer C.free_alloc_viam_carto(pVc)
```
nicksanford, line 157:
```suggestion
	defer C.free_viam_carto(pVc)
```

nicksanford (changes requested):

cartofacade/cgo_api.go
nicksanford: looks great!

cartofacade/cgo_api.go
nicksanford: This works. I realized it would be more complex to also need to carry around the ppVc itself b/c we would also need to remember to free both pointers when we terminate. Your implementation here is best.

cartofacade/cgo_api.go
nicksanford: actually, lets remove the private types which just wrap go types

cartofacade/cgo_api.go
kim-mishra: `alloc_bstring_array takes ``size_t` because malloc requires it. `free_bstring_array` needs to take an int so that we can iterate with it. Is that okay?

cartofacade/cgo_api.go
kim-mishra: removing this type all together

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,456 @@
...
+
+	viam_carto_lib* alloc_viam_carto_lib() { return (viam_carto_lib*) malloc(sizeof(viam_carto_lib)); }
+	void free_alloc_viam_carto_lib(viam_carto_lib* p) { free(p); }
+
+	viam_carto* alloc_viam_carto() { return (viam_carto*) malloc(sizeof(viam_carto)); }
+	void free_alloc_viam_carto(viam_carto* p) { free(p); }
+
+	bstring* alloc_bstring_array(size_t len) { return (bstring*) malloc(len * sizeof(bstring)); }
+	void free_bstring_array(bstring* p, int len) {
+		for (int i = 0; i < len; i++) {
```
nicksanford, line 22: You can iterate with it if you make the following update:

```suggestion
		for (size_t i = 0; i < len; i++) {
```

cartofacade/cgo_api.go
nicksanford: Got it, very cool!

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,456 @@
...
+	case C.VIAM_CARTO_LIB_PLATFORM_INVALID:
+		return errors.New("VIAM_CARTO_LIB_PLATFORM_INVALID")
+	case C.VIAM_CARTO_LIB_INVALID:
+		return errors.New("VIAM_CARTO_LIB_INVALID")
+	case C.VIAM_CARTO_LIB_NOT_INITIALIZED:
+		return errors.New("VIAM_CARTO_LIB_NOT_INITIALIZED")
+	case C.VIAM_CARTO_SENSORS_LIST_EMPTY:
+		return errors.New("VIAM_CARTO_SENSORS_LIST_EMPTY")
+	case C.VIAM_CARTO_UNKNOWN_ERROR:
+		return errors.New("status code unclassified")
```
nicksanford, line 437:
```suggestion
		return errors.New("VIAM_CARTO_UNKNOWN_ERROR")
```

The nuance between this case & the default case, is that this case means that the C++ code threw an exception that didn't have a corresponding status code. The default case means that the C code returned a status code that the go code couldn't classify. We need to be able to distinguish between these 2 cases.

cartofacade/cgo_api.go
```diff
@@ -0,0 +1,456 @@
...
+	case C.VIAM_CARTO_SLAM_MODE_INVALID:
+		return errors.New("VIAM_CARTO_SLAM_MODE_INVALID ")
+	case C.VIAM_CARTO_LIDAR_CONFIG_INVALID:
+		return errors.New("VIAM_CARTO_LIDAR_CONFIG_INVALID ")
+	case C.VIAM_CARTO_MAP_RATE_SEC_INVALID:
+		return errors.New("VIAM_CARTO_MAP_RATE_SEC_INVALID ")
+	case C.VIAM_CARTO_COMPONENT_REFERENCE_INVALID:
+		return errors.New("VIAM_CARTO_COMPONENT_REFERENCE_INVALID ")
+	default:
+		return errors.New("unknown status code")
```
nicksanford, line 449:
```suggestion
		return errors.New("status code unclassified")
```

cartofacade/cgo_api.go
kim-mishra: got it, fixed!

cartofacade/cgo_api_test.go
```diff
@@ -0,0 +1,166 @@
...
+	"time"
+
+	"go.viam.com/test"
+)
+
+func getTestConfig() CartoConfig {
+	return CartoConfig{
+		Sensors:            []string{"rplidar", "imu"},
+		MapRateSecond:      5,
+		DataDir:            "temp",
```
nicksanford, line 15: I meant like `https://pkg.go.dev/os#CreateTemp`

cartofacade/cgo_api_test.go
```diff
@@ -0,0 +1,166 @@
...
+		test.That(t, holder.Real, test.ShouldAlmostEqual, 1100, .001)
+	})
+}
+
+func TestToSensorReading(t *testing.T) {
+	t.Run("sensor reading properly converted between c and go", func(t *testing.T) {
+		timestamp := time.Date(2021, 8, 15, 14, 30, 45, 100, time.Local)
+		sr := toSensorReading([]byte("he0llo"), timestamp)
+		test.That(t, bstringToGoString(sr.sensor_reading), test.ShouldResemble, "he0llo")
+		test.That(t, sr.sensor_reading_time_unix_micro, test.ShouldAlmostEqual, timestamp.UnixMicro(), .001)
```
nicksanford, line 89: This isn't a float, shouldn't this be exactly equal?

cartofacade/cgo_api_test.go
```diff
@@ -0,0 +1,174 @@
...
+	})
+}
+
+func TestGetPositionResponse(t *testing.T) {
+	gpr := getTestGetPositionResponse()
+	t.Run("position response properly converted between C and go", func(t *testing.T) {
+		holder := toGetPositionResponse(gpr)
+		test.That(t, holder.ComponentReference, test.ShouldEqual, "C++ component reference")
+
+		test.That(t, holder.X, test.ShouldAlmostEqual, 100, .001)
```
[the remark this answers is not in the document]
kim-mishra: Nope, now it is passing with `ShouldEqual`, must've had to do with some of the incorrect type conversions you caught! thanks!

nicksanford (approved):

PR #150, opened 2023-06-15
Streamline CMake and build system, move to newer protobuf

Otterverse:
Should be ready for final review. Changes a lot of stuff, but tested a lot so far (but could use more) More notes on testing in comments below.


viam-cartographer/scripts/setup_cartographer.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+if which brew; then
```
jeremyrhyde, line 3: why not include this in makefile?

jeremyrhyde: And if we want to keep it a `.sh` file, should it still be in the scripts folder

viam-cartographer/CMakeLists.txt
```diff
@@ -1,237 +1,81 @@
...
+  ${PCL_LIBRARIES}
+  absl::log_internal_check_op # should be included from protobuf, but isn't
+)
+
+# executables
+add_executable(carto_grpc_server src/main.cc)
+target_link_libraries(carto_grpc_server PUBLIC viam-cartographer)
+
+add_executable(unit_tests ${ALL_VIAM_TEST_SRCS})
+target_link_libraries(unit_tests PUBLIC viam-cartographer)
```
jeremyrhyde, line 81: From 231 lines to 80!!!!

viam-cartographer/scripts/setup_cartographer.sh
```diff
@@ -0,0 +1,20 @@
...
+	echo "Installing cartographer external dependencies"
+	brew update
+	brew install abseil boost ceres-solver protobuf ninja cairo googletest lua@5.3 pkg-config cmake go@1.20 grpc
```
jeremyrhyde, line 6: Waiting on https://github.com/Homebrew/homebrew-core/pull/133818

Makefile
```diff
@@ -38,6 +44,7 @@ ensure-submodule-initialized:
 	else \
 		echo "Submodule found successfully"; \
 	fi
+	cd viam-cartographer/cartographer && git checkout . && git apply ../cartographer_patches/carto.patch
```
jeremyrhyde, line 47: Nice! I like the placement of this patch

jeremyrhyde: What if the ensure submodule command is run twice? Will the second patch cause issues?

Makefile
```diff
@@ -71,26 +78,14 @@ lint-go:
...
-endif
 
-build-debug: build-module
-ifneq ($(wildcard viam-cartographer/cartographer/build/.),)
-	cd viam-cartographer && ./scripts/build_viam_cartographer_debug.sh 
-else 
-	cd viam-cartographer && ./scripts/build_cartographer.sh && ./scripts/build_viam_cartographer_debug.sh
-endif
+build: ensure-submodule-initialized buf build-module
+	cd viam-cartographer && cmake -Bbuild -G Ninja ${EXTRA_CMAKE_FLAGS} && cmake --build build
```
jeremyrhyde, line 85: we dont have to run the `ninja` command anymore?

Makefile
```diff
@@ -71,26 +78,14 @@ lint-go:
...
+
+build-debug: EXTRA_CMAKE_FLAGS+=" -DCMAKE_BUILD_TYPE=Debug"
+build-debug: build
```
jeremyrhyde, line 88: Can we combine these?

viam-cartographer/scripts/setup_cartographer.sh
Otterverse: It is a script... so yeah, it should be in the scripts folder I think. Not sure I understand why it wouldn't be.

And it's not in the makefile because the rest needs to be a script anyway, but this means it can run standalone (without the makefile) too, since the logic can live in bash anyway. (It's also just cleaner to read all in one place.)

Makefile
Otterverse: No, that's why it does a git checkout first... that reverts to the unpatched state. Also, if you make further modifications to ANYTHING in carto, you can quickly save them with just `git diff --output ../cartographer_patches/carto.diff` from within the cartographer submodule directory.

viam-cartographer/scripts/setup_cartographer.sh
jeremyrhyde: I'm fine leaving it in the scripts folder if you think its better there, but we could/should also rename this to just `setup.sh`

Makefile
jeremyrhyde: I had noted on line 82 that we coudl run make commands on separate lines using a @make line first. However, I just ran a test and, no luck.

Otterverse (commented):
Updated the makefile and renamed the setup script.

Otterverse:
Now that protobuf is updated in upstream homebrew, a lot more updates here, but I think it's finally ready for review. All build work on my machine(s) under the following:

canon/apt amd64 (make setup lint build install test) (also appimage)
canon/apt arm64 (make setup lint build install test) (also appimage)
linux/brew amd64 (make setup lint build install test)
mac/brew arm64 (make setup lint build install test)

The github workflows that were changed were tested manually:
test.yml https://github.com/viamrobotics/viam-cartographer/actions/runs/5327906791
appimage.yml https://github.com/viamrobotics/viam-cartographer/actions/runs/5327704981


That said, please test further on your own team and make sure I haven't broken some use case or workflow I wasn't aware of.

Makefile
Otterverse: I was mistaken about suppression. The @ symbol only suppresses the echoing of the command, not the output. That said, I still think it's better to use true dependencies rather than sub-make processes (which are effectively just calling make again from a new shell.)

viam-cartographer/cartographer_patches/carto.patch
```diff
@@ -0,0 +1,304 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
```
Otterverse, line 1: This file gets applied on top of the upstream cartographer submodule during ensure-submodule-initialization. Changes within that submodule can thus be made "permanent" with `git diff --output ...` set to this file.

viam-cartographer/cartographer_patches/carto.patch
```diff
@@ -0,0 +1,304 @@
...
+-      DEPENDS ${ABS_FIL}
+-      COMMENT "Running C++ protocol buffer compiler on ${ABS_FIL}"
+-      VERBATIM
+-    )
+-  endforeach()
+-  set_source_files_properties(${ALL_GRPC_SERVICE_SRCS} ${ALL_GRPC_SERVICE_HDRS} PROPERTIES GENERATED TRUE)
+-  list(APPEND ALL_LIBRARY_HDRS ${ALL_GRPC_SERVICE_HDRS})
+-  list(APPEND ALL_LIBRARY_SRCS ${ALL_GRPC_SERVICE_SRCS})
+-endif()
+ set(INSTALL_GENERATED_HDRS ${ALL_PROTO_HDRS} ${ALL_GRPC_SERVICE_HDRS})
```
jeremyrhyde, line 177: [q] is this needed after all other references to ALL_GRPC_SERVICE_HDRS have been deleted?

viam-cartographer/cartographer_patches/carto.patch
```diff
@@ -0,0 +1,304 @@
...
+ 
+ target_include_directories(${PROJECT_NAME} SYSTEM PUBLIC
+@@ -280,60 +180,19 @@ target_link_libraries(${PROJECT_NAME} PUBLIC ${PROTOBUF_LIBRARY}
+   absl::strings
+   absl::synchronization
+   absl::time
+-  absl::utility 
++  absl::utility
+ )
++if (TARGET absl::log_internal_check_op)
```
jeremyrhyde, line 234: [note to self] This doesn't exist in protobuf < 23.0 so we need to case on it

viam-cartographer/cartographer_patches/carto.patch
```diff
@@ -0,0 +1,304 @@
...
+diff --git a/cmake/functions.cmake b/cmake/functions.cmake
+index 8cfc05f..6d849ca 100644
+--- a/cmake/functions.cmake
++++ b/cmake/functions.cmake
+@@ -87,7 +87,7 @@ macro(google_initialize_cartographer_project)
+     google_add_flag(GOOG_CXX_FLAGS "-Werror=reorder")
+     google_add_flag(GOOG_CXX_FLAGS "-Werror=return-type")
+     google_add_flag(GOOG_CXX_FLAGS "-Werror=switch")
+-    google_add_flag(GOOG_CXX_FLAGS "-Werror=uninitialized")
++    # google_add_flag(GOOG_CXX_FLAGS "-Werror=uninitialized")
```
jeremyrhyde, line 301: [q] why is this being removed?

etc/packaging/appimages/cartographer-module-x86_64.yml
```diff
@@ -7,13 +7,10 @@ script:
...
  - cp ../../../bin/cartographer-module $TARGET_APPDIR/usr/bin/
  - cp ../../../viam-cartographer/build/carto_grpc_server $TARGET_APPDIR/usr/bin/
- - cp ../../../viam-cartographer/cartographer/build/libcartographer.a $TARGET_APPDIR/usr/lib/
- - cp ../../../viam-cartographer/build/libviam-cartographer.a $TARGET_APPDIR/usr/lib/
  - cp ../../../viam-cartographer/lua_files/* $TARGET_APPDIR/usr/share/cartographer/lua_files
  - mkdir -p $TARGET_APPDIR/usr/share/icons/viam/256x256/apps/
  - cp ./viam-server.png $TARGET_APPDIR/usr/share/icons/viam/256x256/apps/viam-server.png
- - go install github.com/Otterverse/aix@latest
- - cp `go env GOPATH`/bin/aix $TARGET_APPDIR/usr/bin/
+ - GOBIN=$TARGET_APPDIR/usr/bin/ go install github.com/Otterverse/aix@latest
```
jeremyrhyde, line 13: [q] why isnt this change in rdk?

viam-cartographer/CMakeLists.txt
```diff
@@ -1,237 +1,80 @@
...
 project(viam-cartographer)
-add_definitions(-DABSL_LEGACY_THREAD_ANNOTATIONS)
+set (CMAKE_CXX_STANDARD 17)
 
-set(VIAM_CARTOGRAPHER_MAJOR_VERSION 1)
-set(VIAM_CARTOGRAPHER_MINOR_VERSION 0)
-set(VIAM_CARTOGRAPHER_PATCH_VERSION 0)
-set(VIAM_CARTOGRAPHER_VERSION ${CARTOGRAPHER_MAJOR_VERSION}.${CARTOGRAPHER_MINOR_VERSION}.${CARTOGRAPHER_PATCH_VERSION})
-set(VIAM_CARTOGRAPHER_SOVERSION ${CARTOGRAPHER_MAJOR_VERSION}.${CARTOGRAPHER_MINOR_VERSION})
+if(DEFINED ENV{API_SOURCE_DIR})
```
jeremyrhyde, line 6: [q] where is this defined?

Makefile
```diff
@@ -1,35 +1,44 @@
...
 clean:
-	rm -rf grpc
-	rm -rf bin
-	rm -rf viam-cartographer/build
-	rm -rf viam-cartographer/cartographer/build
+	rm -rf grpc bin viam-cartographer/build
 
 clean-all:
 	git clean -fxd
+	cd viam-cartographer/cartographer && git checkout . && git clean -fxd
```
jeremyrhyde, line 41: [note to self] resets CMakeList

Makefile
```diff
@@ -38,21 +47,8 @@ ensure-submodule-initialized:
...
-endif
-
-lint-setup-go:
-	GOBIN=`pwd`/$(TOOL_BIN) go install \
-		github.com/edaniels/golinters/cmd/combined \
-		github.com/golangci/golangci-lint/cmd/golangci-lint \
-		github.com/rhysd/actionlint/cmd/actionlint
-
-lint-setup: lint-setup-cpp lint-setup-go
+	grep -q viam-patched viam-cartographer/cartographer/CMakeLists.txt || \
```
jeremyrhyde, line 50: [note to self] now when submodule is initialized, CMakeList will auto be applied

Makefile
```diff
@@ -1,35 +1,44 @@
...
 
-bufinstall:
-	sudo apt-get install -y protobuf-compiler-grpc libgrpc-dev libgrpc++-dev || brew install grpc openssl --quiet
+$(TOOL_BIN)/buf:
+	go install github.com/bufbuild/buf/cmd/buf@v1.8.0
 
-bufsetup:
-	GOBIN=`pwd`/grpc/bin go install github.com/bufbuild/buf/cmd/buf@v1.8.0
-	ln -sf `which grpc_cpp_plugin` grpc/bin/protoc-gen-grpc-cpp
+$(TOOL_BIN)/protoc-gen-grpc-cpp:
```
jeremyrhyde, line 26: [q] what is the benefit of moving to this more complex make target name?

Makefile
```diff
@@ -63,42 +59,46 @@ lint-cpp:
...
 
 lint: ensure-submodule-initialized lint-cpp lint-go
 
-setup: ensure-submodule-initialized
-ifeq ("Darwin", "$(shell uname -s)")
-	cd viam-cartographer/scripts && ./setup_cartographer_macos.sh
+setup: install-dependencies ensure-submodule-initialized artifact-pull
+
+install-dependencies:
+ifneq (, $(shell which brew))
```
jeremyrhyde, line 78: [note to self] from setup scripts

Makefile
```diff
@@ -1,35 +1,44 @@
...
+TOOL_BIN := $(shell pwd)/bin/tools/$(shell uname -s)-$(shell uname -m)
+GIT_REVISION := $(shell git rev-parse HEAD | tr -d '\n')
+TAG_VERSION ?= $(shell git tag --points-at | sort -Vr | head -n1)
+GO_BUILD_LDFLAGS := -ldflags "-X 'main.Version=${TAG_VERSION}' -X 'main.GitRevision=${GIT_REVISION}'"
+SHELL := /usr/bin/env bash
+export PATH := $(TOOL_BIN):$(PATH)
+export GOBIN := $(TOOL_BIN)
+
+ifneq (, $(shell which brew))
+	EXTRA_CMAKE_FLAGS := -DCMAKE_PREFIX_PATH=$(shell brew --prefix) -DQt5_DIR=$(shell brew --prefix qt5)/lib/cmake/Qt5
```
jeremyrhyde, line 11: [note to self] from build scripts

.github/workflows/test.yml
```diff
@@ -57,59 +57,47 @@ jobs:
...
+        apt-get update
 
     - name: Verify no uncommitted changes from make lint
       run: |
         git init
         git add .
         chown -R testbot:testbot .
-        sudo -u testbot bash -lc 'make lint-setup lint'
-        GEN_DIFF=$(git status -s)
+        sudo -u testbot bash -lc 'sudo apt-get install -y clang-format && make lint'
```
jeremyrhyde, line 67: [q] Did you remove the brew install as these ymls are only run in linux based CI?

.github/workflows/appimage.yml
```diff
@@ -48,31 +46,20 @@ jobs:
...
-    - name: make build
+    - name: make setup
       run: |
-        sudo -u testbot bash -lc 'cd viam-cartographer && make build'
+        sudo -u testbot bash -lc 'make setup'
 
-    - name: make install-lua-files
+    - name: make build
       run: |
-        sudo -u testbot bash -lc 'cd viam-cartographer && make install-lua-files'
```
jeremyrhyde, line 70: [q] Why isn't `install-lua-files` needed anymore? Was it never needed during Appimage builds?

.github/workflows/appimage.yml
```diff
@@ -48,31 +46,20 @@ jobs:
     - name: make clean
       run: |
         chown -R testbot:testbot .
-        sudo -u testbot bash -lc 'cd viam-cartographer && make clean'
```
jeremyrhyde, line 51: [q] Is removing this a benefit of the (TOOL_BIN) additions made to the Makefile?

cartofacade/cgo_api.go
```diff
@@ -8,7 +8,7 @@ package cartofacade
 	#cgo CFLAGS: -I../viam-cartographer/src/carto_facade
 
 	// the libraries that need to be linked can be derived from line 258 of the build.ninja file that is autogenerated during make build
-	#cgo LDFLAGS: -L../viam-cartographer/build -L../viam-cartographer/cartographer/build -lviam-cartographer  -lcartographer -ldl -lm -labsl_hash  -labsl_city -labsl_bad_optional_access -labsl_strerror  -labsl_str_format_internal -labsl_synchronization -labsl_strings -labsl_throw_delegate -lcairo -llua5.3 -lstdc++ -lceres -lprotobuf -lglog -lboost_filesystem -lboost_iostreams -lpcl_io -lpcl_common -labsl_raw_hash_set
+	#cgo LDFLAGS: -L../viam-cartographer/build -L../viam-cartographer/build/cartographer -lviam-cartographer  -lcartographer -ldl -lm -labsl_hash  -labsl_city -labsl_bad_optional_access -labsl_strerror  -labsl_str_format_internal -labsl_synchronization -labsl_strings -labsl_throw_delegate -lcairo -llua5.3 -lstdc++ -lceres -lprotobuf -lglog -lboost_filesystem -lboost_iostreams -lpcl_io -lpcl_common -labsl_raw_hash_set
```
jeremyrhyde, line 11: [q] Why did the build and cartographer swap places?

etc/packaging/appimages/cartographer-module-x86_64.yml
Otterverse: Because I wrote the RDK stuff a year and a half ago before I knew about GOBIN. :-)

(Also, it still works there... but this had a conflict with the "old" way, as GOBIN was being set to TOOL_BIN in the makefile, go the default location was no good anymore.)

viam-cartographer/CMakeLists.txt
Otterverse: It'd be something you set on the command line typically `API_SOURCE_DIR=/tmp/source/api make build` or with a -D option for cmake `cmake -DAPI_SOURCE_DIR=/tmp/source/api`

Also, I just moved this from lower down in the file (didn't create it) as I was rearranging things to a more logical flow.

viam-cartographer/cartographer_patches/carto.patch
```diff
@@ -0,0 +1,304 @@
...
+-endif()
+-
+ # Install catkin package.xml
+ install(FILES package.xml DESTINATION share/cartographer)
+ 
+@@ -80,9 +51,6 @@ install(DIRECTORY cmake DESTINATION share/cartographer/)
+ 
+ file(GLOB_RECURSE ALL_LIBRARY_HDRS "cartographer/*.h")
+ file(GLOB_RECURSE ALL_LIBRARY_SRCS "cartographer/*.cc")
+-file(GLOB_RECURSE TEST_LIBRARY_HDRS "cartographer/fake_*.h" "cartographer/*test_helpers*.h" "cartographer/mock_*.h")
```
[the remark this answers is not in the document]
Otterverse: I don't know. It wasn't my change. 90% of this file was already here. I just turned it from a file that got copied on top of the submodule, to a patch against it, since I also needed to modify another file.

A patch is just easier, as it can be any number of files/changes, vs having to explicitly copy multiple files to specific locations.

The only real changes I made to the inner CMakeLists.txt were removing the CERES_INCLUDE_DIRS bit, adding the comment at the top, and adding the conditional link libraries for absl::log_internal_check_op

viam-cartographer/cartographer_patches/carto.patch
```diff
@@ -0,0 +1,304 @@
...
++
++# Prometheus files are part of the cartographer library - but we don't need them.
++list(REMOVE_ITEM ALL_LIBRARY_HDRS ${ALL_PROMETHEUS_FILES})
++list(REMOVE_ITEM ALL_LIBRARY_SRCS ${ALL_PROMETHEUS_FILES})
++list(REMOVE_ITEM TEST_LIBRARY_HDRS ${ALL_PROMETHEUS_FILES})
++list(REMOVE_ITEM TEST_LIBRARY_SRCS ${ALL_PROMETHEUS_FILES})
++list(REMOVE_ITEM ALL_TESTS ${ALL_PROMETHEUS_FILES})
++list(REMOVE_ITEM ALL_EXECUTABLES ${ALL_PROMETHEUS_FILES})
++
+ set(INSTALL_SOURCE_HDRS ${ALL_LIBRARY_HDRS} ${TEST_LIBRARY_HDRS})
```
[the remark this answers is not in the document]
Otterverse: Per previous comment... I don't know. Most of these changes weren't mine, and I wanted to focus just on cleaning up the outer layer for now.

viam-cartographer/cartographer_patches/carto.patch
Otterverse: Per previous, not my changes, so can't say for sure. I suspect it's not though. 

Cleaning this up could be good for a followup task on your team. You can checkout the original version just add a more minimal or cleaner set of changes.

viam-cartographer/cartographer_patches/carto.patch
Otterverse: More specifically, it didn't exist in some previous version of abseil. But that also means protobuf didn't USE it either. And thus this is a crude check... if it exists, assume protobuf needs it, as both packages probably got updated at the same time.

viam-cartographer/cartographer_patches/carto.patch
Otterverse: -Werror turns a warning into a (fatal) error. Newer GCC finds uninitialized values here that older GCC didn't, so this causes the entire build to fail on newer linux distros. With this disabled, they become just warnings again.

Otterverse (commented):
Okay, I think that's all the questions answered. Feel free to add more as you find them though.

Makefile
JohnN193: gotcha, so thats why other repos were updated to use `dist/buf` and similar setups

viam-cartographer/CMakeLists.txt
JohnN193: yeah we added this forever ago when protos still lived in rdk. Since the `build` step now depends on `buf` we can probably remove this at some point

JohnN193 (approved):
LGTM, just a few questions I had. thanks for doing this!

.github/workflows/appimage.yml
```diff
@@ -48,31 +46,20 @@ jobs:
...
+        sudo -u testbot bash -lc 'make clean'
         
-    - name: make bufinstall buf setup
-      uses: nick-fields/retry@v2
-      with:
-        timeout_minutes: 5
-        max_attempts: 3
-        retry_on: timeout
```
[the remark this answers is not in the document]
Otterverse: I haven't seen it be an issue in a while, which is why I removed it. If it is an issue still (and it starts happening again) we should correct the source of it, which might mean shifting this to the buildjet runners if needed. Good note though!

Makefile
```diff
@@ -1,35 +1,44 @@
...
+	export PKG_CONFIG_PATH := $(shell brew --prefix openssl@3)/lib/pkgconfig
```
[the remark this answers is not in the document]
Otterverse: Yes, it would. But I don't think that's an issue as it would only be for the duration of this command, AND we verified that brew is installed and in use. If someone is trying to hack in openssl (or other custom packages) from somewhere else, then they can/should know enough to modify this file themselves. :-)

Makefile
Otterverse: Yeah, though in that case, it was using a slightly different pattern where the "dist" folder contains touchfiles (just empty files with specific names) to basically flag when a job had been "done." You can then reset everything by just wiping those files out, causing a clean build to happen. Sometimes when you're linking in more complex tooling (that might have dozens of things it generates, like the buf generate command does) that's an easier way than setting up targets for EVERY output file. Especially if the output files often change or are variable in number.

Makefile
Otterverse: Yeah, in theory you can just change -Bbuild to -Bbuild-\`uname -m\`  (or whatever name scheme you want, and it'll configure the output there. Then you can actually compile with the same thing just changing build to build-\`uname -m\` in the final command as well.

More realistically, you'd just want to make them variables in Make, then you can set the BUILD_DIR variable once and use everywhere. Do note that some of the cartographer tests in go are written using cgo and depending on explicit paths (which is why I had to change one) so there's not as easy of a way to handle that.

Likewise, it's not just the cmake part that would need to be per-architecture, as the golang stuff would be as well, so it's likely a bit of a more complex change you really want, to completely split out architectures if you really want to be able to build both in the same code space (without just doing "make clean" instead.)

viam-cartographer/CMakeLists.txt
Otterverse: The reason you _might_ want to keep it (and why I left it) is when you're testing NEW api changes. This easily lets you use something local.

nicksanford:
@Otterverse I think the lint step is broken on this PR

viam-cartographer/cartographer_patches/carto.patch
jeremyrhyde: Should this line just be fully removed then?

jeremyrhyde (approved):
LGTM, thanks for answering all my questions James!

nicksanford (commented):
First off:
1. Thank you so much for this update @Otterverse ! I learned a lot about make and cmake from your PR.
2. I've left a few questions

.github/workflows/test.yml
```diff
@@ -57,59 +57,47 @@ jobs:
...
+        GEN_DIFF=$(git status -s --ignore-submodules)
```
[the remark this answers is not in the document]
Otterverse: Because we're patching it during ensure-submodule-initalized (which the "lint" target depends on.)

Makefile
```diff
@@ -1,35 +1,44 @@
...
-TOOL_BIN = bin/gotools/$(shell uname -s)-$(shell uname -m)
-PATH_WITH_TOOLS="`pwd`/$(TOOL_BIN):$(HOME)/go/bin/:${PATH}"
-GIT_REVISION = $(shell git rev-parse HEAD | tr -d '\n')
-TAG_VERSION?=$(shell git tag --points-at | sort -Vr | head -n1)
-GO_BUILD_LDFLAGS = -ldflags "-X 'main.Version=${TAG_VERSION}' -X 'main.GitRevision=${GIT_REVISION}'"
+TOOL_BIN := $(shell pwd)/bin/tools/$(shell uname -s)-$(shell uname -m)
+GIT_REVISION := $(shell git rev-parse HEAD | tr -d '\n')
+TAG_VERSION ?= $(shell git tag --points-at | sort -Vr | head -n1)
+GO_BUILD_LDFLAGS := -ldflags "-X 'main.Version=${TAG_VERSION}' -X 'main.GitRevision=${GIT_REVISION}'"
+SHELL := /usr/bin/env bash
```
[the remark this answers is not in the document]
Otterverse: There's weird behavior between different OSes with different shells. Hit it as an actual bug which was fixed here: https://github.com/viamrobotics/api/pull/267 That PR has some more details on the issue.

Makefile
```diff
@@ -1,35 +1,44 @@
...
+$(TOOL_BIN)/protoc-gen-grpc-cpp:
+	mkdir -p "$(TOOL_BIN)"
+	which grpc_cpp_plugin && ln -sf `which grpc_cpp_plugin` $(TOOL_BIN)/protoc-gen-grpc-cpp
 
-buf: bufsetup
-	PATH="${PATH}:`pwd`/grpc/bin" buf generate --template ./buf/buf.gen.yaml buf.build/viamrobotics/api
-	PATH="${PATH}:`pwd`/grpc/bin" buf generate --template ./buf/buf.grpc.gen.yaml buf.build/viamrobotics/api
-	PATH="${PATH}:`pwd`/grpc/bin" buf generate --template ./buf/buf.gen.yaml buf.build/googleapis/googleapis
+buf: $(TOOL_BIN)/buf $(TOOL_BIN)/protoc-gen-grpc-cpp
+	echo ${PATH}
```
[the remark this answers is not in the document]
Otterverse: Doh! Good catch.

Makefile
```diff
@@ -63,42 +59,46 @@ lint-cpp:
...
 
-lint-go:
+lint-go: $(TOOL_BIN)/combined $(TOOL_BIN)/golangci-lint $(TOOL_BIN)/actionlint
 	go vet -vettool=$(TOOL_BIN)/combined ./...
-	GOGC=50 $(TOOL_BIN)/golangci-lint run -v --fix --config=./etc/golangci.yaml
-	PATH=$(PATH_WITH_TOOLS) actionlint
+	GOGC=50 golangci-lint run -v --fix --config=./etc/golangci.yaml
+	actionlint
+
+$(TOOL_BIN)/combined $(TOOL_BIN)/golangci-lint $(TOOL_BIN)/actionlint:
```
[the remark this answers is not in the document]
Otterverse: It runs the target if any of them are missing (or out of date.)

viam-cartographer/CMakeLists.txt
```diff
@@ -1,237 +1,80 @@
...
-include_directories(${PROJECT_SOURCE_DIR}/cartographer)
-include_directories(${PROJECT_SOURCE_DIR}/cartographer/build)
-link_libraries(${PROJECT_SOURCE_DIR}/cartographer/build/libcartographer.a)
 
-find_package(PCL REQUIRED)
-include_directories(${PCL_INCLUDE_DIRS})
-link_directories(${PCL_LIBRARY_DIRS})
-add_definitions(${PCL_DEFINITIONS})
+# dependencies
+add_subdirectory(cartographer)
```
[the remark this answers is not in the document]
Otterverse: include_directories() adds locations to pass to the compiler when it's looking for header files. (E.g. stuff passed with -L)

add_subdirectory() adds a CMake directory with it's own CMakeLists.txt in it and "imports" it, so you can use those definitions/targets/etc. at the CMake level.

viam-cartographer/CMakeLists.txt
```diff
@@ -1,237 +1,80 @@
...
 
-find_package(OpenSSL REQUIRED)
+find_package(PkgConfig REQUIRED)
+pkg_check_modules(GRPCPP REQUIRED grpc++)
```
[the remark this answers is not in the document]
Otterverse: No, because we're not directly using it or the variables it exports in CMake. The inner "cartographer" target already depends on it, and we depend on it in turn, that's sufficient as far as I'm aware.

viam-cartographer/CMakeLists.txt
```diff
@@ -1,237 +1,80 @@
...
-    absl::meta
-    absl::numeric
-    absl::str_format
-    absl::strings
-    absl::synchronization
-    absl::time
-    absl::utility 
-  )
-  if (NOT WIN32)
-    target_link_libraries(${PROJECT_NAME} PUBLIC pthread)
```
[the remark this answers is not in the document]
Otterverse: Like with protobuf, this outer project ("viam-cartographer") doesn't directly use it. The inner "cartographer" target does (and has this.) So simply depending on "cartographer" (through target_link_libraries()) is enough.

Otterverse (commented):
Answered nick's questions, and removed that stray echo. Also rebased on main. I think it's good hopefully.

nicksanford (approved):

nicksanford:
@Otterverse this looks great to me! Thanks for all the work you put in & for making the cartographer build system much more reliable & maintainable.

PR #158, opened 2023-06-23
update-makefile-dont-buf-every-time

nicksanford:
I noticed after the update to streamline CMake & the Makefile that running any make command that depends on `buf` runs `make buf` every time, which results in all the GRPC C++ classes being compiled every time, which dramatically slows down the workflow of incremental C++ builds (when doing TDD for example).

This PR makes it so that `buf` (now called `grpc/buf` will touch a file called `grpc/buf` once it succeeds which should make it so that this commend doesn't continue to get run once it succeeds once.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,10 +27,16 @@ $(TOOL_BIN)/protoc-gen-grpc-cpp:
 	mkdir -p "$(TOOL_BIN)"
 	which grpc_cpp_plugin && ln -sf `which grpc_cpp_plugin` $(TOOL_BIN)/protoc-gen-grpc-cpp
 
-buf: $(TOOL_BIN)/buf $(TOOL_BIN)/protoc-gen-grpc-cpp
+grpc/buf: $(TOOL_BIN)/buf $(TOOL_BIN)/protoc-gen-grpc-cpp
 	buf generate --template ./buf/buf.gen.yaml buf.build/viamrobotics/api
 	buf generate --template ./buf/buf.grpc.gen.yaml buf.build/viamrobotics/api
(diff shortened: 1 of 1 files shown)


Otterverse (commented):
Changes look fine, but just want to confirm there isn't something else going on.

When I originally switched this, it was to make sure things never got built with out-of-date generated files, which has/had been a big problem in the past. In timing it, `make buf` (on a second run) takes 0.82 seconds on my desktop, and 1.35 seconds on my macbook. You said it dramatically slows things, so I'm curious if you're seeing different behavior than I am, or if you're just doing something different where 1.4 seconds is too long. I'd definitely assumed a 1% slowdown was acceptable.

nicksanford:
> Changes look fine, but just want to confirm there isn't something else going on.
> 
> When I originally switched this, it was to make sure things never got built with out-of-date generated files, which has/had been a big problem in the past. In timing it, `make buf` (on a second run) takes 0.82 seconds on my desktop, and 1.35 seconds on my macbook. You said it dramatically slows things, so I'm curious if you're seeing different behavior than I am, or if you're just doing something different where 1.4 seconds is too long. I'd definitely assumed a 1% slowdown was acceptable.

Its not that the `buf` step is slow. It's that it generates C++ source files that then need to be recompiled, even if the file contents have not changed. I'll take timings on how much time that takes & get back to you.

Otterverse (approved):
Never mind my previous comment. I see it now! It's not that the buf itself is slow, it's that the files it generates get their timestamps bumped, so have to be recompiled as well, meaning there's always at least 17ish files to build, even on a no-change. Got it!

Otterverse:
Jinx! :-)

PR #159, opened 2023-06-23
RSDK-3561: Implement GetLatestMapInfo in viam-cartographer

PStrutz:
JIRA Ticket: https://viam.atlassian.net/browse/RSDK-3561

viam-cartographer implementation of GetLatestMapInfo;
added mapTimestamp field to cartographerService;
keep track of changing maps when GetPointCloudMap is called


viam-cartographer.go
```diff
@@ -244,6 +246,23 @@ func (cartoSvc *cartographerService) GetInternalState(ctx context.Context) (func
 	return grpchelper.GetInternalStateCallback(ctx, cartoSvc.Name().ShortName(), cartoSvc.clientAlgo)
 }
 
+// GetLatestMapInfo forwards the request for timestamp data to the slam library's gRPC service.
+// Once a response is received, it is converted into a time.Time object.
```
jeremyrhyde, line 250: [major] The way this function is currently written, it will forward this call to the slam algorithm's grpc server (in this case this server is being run by the c++ binary). However, for GetLatestMapInfo, we want to instead directly return the time recorded in GetPointCloudMap.

viam-cartographer.go
```diff
@@ -231,7 +233,7 @@ func (cartoSvc *cartographerService) GetPosition(ctx context.Context) (spatialma
 func (cartoSvc *cartographerService) GetPointCloudMap(ctx context.Context) (func() ([]byte, error), error) {
 	ctx, span := trace.StartSpan(ctx, "viamcartographer::cartographerService::GetPointCloudMap")
 	defer span.End()
-
+	cartoSvc.mapTimestamp = time.Now()
```
jeremyrhyde, line 236: [major] I believe you are missing some logic regarding when to update the timestamp/call a new map when running in localization mode. Think about the two flows:
```
LOCALIZATION
Startup: timestamp = timestamp1
First call to GetProperties: timestamp = timestamp1
Timestamp has changed: Call GetMap: return map, no update to timestamp
Second call to GetProperties: timestamp = timestamp1
Timestamp has not changed: no call to GetMap
....
Reconfiguration occurs  timestamp = timestamp2
Nth call to GetProperties:timestamp = timestamp2
Timestamp has changed: Call  GetMap: return map, no update to timestamp
N+1th call to GetProperties:timestamp = timestamp2
Timestamp has not changed: no call to GetMap
NON-LOCALIZATION
Startup: timestamp = timestamp1
First call to GetProperties: timestamp = timestamp1
Timestamp has changed: Call GetMap: return map, update timestamp = timestamp2
Second call to GetProperties: timestamp = timestamp2
Timestamp has changed: Call GetMap: return map, update timestamp = timestamp3
....
Reconfiguration occurs  timestamp = timestamp_n
Nth call to GetProperties:timestamp = timestamp_n
Timestamp has changed: Call  GetMap: return map, update timestamp = timestamp_n+1
N+1th call to GetProperties:timestamp = timestamp_n+1
Timestamp has not changed: Call  GetMap: return map, update timestamp = timestamp_n+2
```

viam-cartographer.go
```diff
@@ -244,6 +246,23 @@ func (cartoSvc *cartographerService) GetInternalState(ctx context.Context) (func
...
+	if err != nil {
+		return time.Time{}, errors.Wrap(err, "error getting latest map info")
+	}
+	latestMapTimestamp := resp.LastMapUpdate.AsTime()
+
+	return latestMapTimestamp, err
+}
+
 // StartDataProcess starts a go routine that saves data from the lidar to the user-defined data directory.
 func (cartoSvc *cartographerService) StartDataProcess(
```
jeremyrhyde, line 267: [note] still need to add reconfigure logic

jeremyrhyde (changes requested):
Review the logic chain for both localization and non-localization mode. Also start thinking of how you would add in some automated tests for this? hint: look at integration.go

viam-cartographer.go
```diff
@@ -142,6 +142,8 @@ func New(
 		cancelFunc:            cancelFunc,
 		logger:                logger,
 		bufferSLAMProcessLogs: bufferSLAMProcessLogs,
+		inLocalization:        mapRateSec == 0,
```
jeremyrhyde, line 145: [nit/opt] localizationMode

viam-cartographer.go
```diff
@@ -228,10 +232,14 @@ func (cartoSvc *cartographerService) GetPosition(ctx context.Context) (spatialma
 
 // GetPointCloudMap creates a request, calls the slam algorithms GetPointCloudMap endpoint and returns a callback
 // function which will return the next chunk of the current pointcloud map.
+// If startup is in localization mode, the timestamp is NOT updated.
```
jeremyrhyde, line 235: Great comment add. Maybe edit the original comment to be something like:
```
// GetPointCloudMap creates a request, **recording the time**, calls the slam algorithms GetPointCloudMap endpoint and returns a callback
// function which will return the next chunk of the current pointcloud map.
```

viam-cartographer.go
```diff
@@ -325,6 +341,13 @@ func (cartoSvc *cartographerService) getNextDataPoint(ctx context.Context, lidar
 	}
 }
 
+// Reconfigure always returns a must rebuild error.
+// Reconfigure requires map update, so map timestamp is updated.
```
jeremyrhyde, line 345: [nit/opt] combine comments

viam-cartographer.go
```diff
@@ -325,6 +341,13 @@ func (cartoSvc *cartographerService) getNextDataPoint(ctx context.Context, lidar
...
+func (cartoSvc *cartographerService) Reconfigure(ctx context.Context, deps resource.Dependencies, conf resource.Config) error {
+	cartoSvc.mapTimestamp = time.Now()
```
jeremyrhyde, line 347: Very clean implementation!!!!

jeremyrhyde (commented):
LGTM, will hold off approving until tests are in but other than a few small nitpicks and opinions (those are the [nit]s and [opt]s) it looks great!

internal/inject/slam_service.go
```diff
@@ -53,6 +55,16 @@ func (slamSvcClient *SLAMServiceClient) GetInternalState(ctx context.Context, in
...
+func (slamSvcClient *SLAMServiceClient) GetLatestMapInfo(ctx context.Context, in *v1.GetLatestMapInfoRequest, opts ...grpc.CallOption) (
+	*v1.GetLatestMapInfoResponse, error,
+) {
+	if slamSvcClient.GetLatestMapInfoFunc == nil {
+		return nil, errors.New("no GetLatestMapInfoFunc defined for injected SLAM service client")
+	}
+	return slamSvcClient.GetLatestMapInfoFunc(ctx, in, opts...)
+}
+
 // DoCommand calls the injected DoCommand or the real variant.
```
jeremyrhyde, line 68: [minor] all the comments in here need to be updated to 
`// XXXX calls the injected XXXXXX if defined else it will return an error`
This is a reminder of the testutils code in rdk which can use real versions.

viam-cartographer.go
```diff
@@ -244,6 +252,14 @@ func (cartoSvc *cartographerService) GetInternalState(ctx context.Context) (func
 	return grpchelper.GetInternalStateCallback(ctx, cartoSvc.Name().ShortName(), cartoSvc.clientAlgo)
 }
 
+// GetLatestMapInfo returns the timestamp stored when GetPointCloudMap was last called.
```
jeremyrhyde, line 255: [nit] slight more descriptive
```suggestion
// GetLatestMapInfo returns the timestamp associated with the latest call to GetPointCloudMap, unless you are localizing; in which case the timestamp returned is the timestamp of the session.
```

jeremyrhyde: Something like that

viam-cartographer.go
```diff
@@ -325,6 +341,13 @@ func (cartoSvc *cartographerService) getNextDataPoint(ctx context.Context, lidar
 	}
 }
 
+// Reconfigure always returns a must rebuild error and requires a map update, so the
```
jeremyrhyde, line 344:
```suggestion
// Reconfigure always returns a must rebuild error and updates the stored timestamp to correspond to the beginning of the new session.
```

viam-cartographer.go
```diff
@@ -325,6 +341,13 @@ func (cartoSvc *cartographerService) getNextDataPoint(ctx context.Context, lidar
...
+// map timestamp is updated.
+func (cartoSvc *cartographerService) Reconfigure(ctx context.Context, deps resource.Dependencies, conf resource.Config) error {
+	cartoSvc.mapTimestamp = time.Now()
```
jeremyrhyde, line 347: [q] did you look at all into the behavior of reconfigure and if a return of `resource.NewMustRebuildError` will call New again?

viam-cartographer_test.go
```diff
@@ -34,6 +36,10 @@ var (
 	testMapRateSec = 200
 	_true          = true
 	_false         = false
+	timestamp1     = time.Time{}
```
jeremyrhyde, line 39: remove timestamp1/2 as they are always defined by the function. _zeroInt and _zeroTime can remain as they are valid comparators

viam-cartographer_test.go
```diff
@@ -356,6 +409,30 @@ func TestSLAMProcess(t *testing.T) {
...
+			UseLiveData:   &_false,
+		}
+
+		svc, err := testhelper.CreateSLAMService(t, attrCfg, logger, false, testExecutableName)
+		test.That(t, err, test.ShouldBeNil)
+
+		timestamp1, _ = svc.GetLatestMapInfo(context.Background())
+		svc.Reconfigure(context.Background(), resource.Dependencies{}, resource.Config{})
+		timestamp2, _ = svc.GetLatestMapInfo(context.Background())
+		test.That(t, timestamp1, test.ShouldNotEqual, &_zeroTime)
```
jeremyrhyde, line 428: You can simply these by using a test.ShouldBeBetween

viam-cartographer_test.go
```diff
@@ -127,6 +133,53 @@ func TestNew(t *testing.T) {
...
+
+		svc, err := testhelper.CreateSLAMService(t, attrCfg, logger, false, testExecutableName)
+		test.That(t, err, test.ShouldBeNil)
+
+		grpcServer.Stop()
+		timestamp1, _ = svc.GetLatestMapInfo(context.Background())
+		svc.GetPointCloudMap(context.Background())
+		timestamp2, _ = svc.GetLatestMapInfo(context.Background())
+		test.That(t, timestamp1, test.ShouldNotEqual, &_zeroTime)
+		test.That(t, timestamp2, test.ShouldNotEqual, &_zeroTime)
```
jeremyrhyde, line 179: see comment below

viam-cartographer_test.go
```diff
@@ -127,6 +133,53 @@ func TestNew(t *testing.T) {
...
+			Port:          "localhost:" + strconv.Itoa(port),
+			UseLiveData:   &_false,
+			MapRateSec:    &_zeroInt,
+		}
+
+		svc, err := testhelper.CreateSLAMService(t, attrCfg, logger, false, testExecutableName)
+		test.That(t, err, test.ShouldBeNil)
+
+		grpcServer.Stop()
+		timestamp1, _ = svc.GetLatestMapInfo(context.Background())
```
jeremyrhyde, line 153: add in the nil error check here. Even though it appears redundant, its worth doing incase the behavior changes

viam-cartographer_test.go
```diff
@@ -356,6 +409,30 @@ func TestSLAMProcess(t *testing.T) {
...
+			DataDirectory: dataDir,
+			Port:          "localhost:" + strconv.Itoa(port),
+			UseLiveData:   &_false,
+		}
+
+		svc, err := testhelper.CreateSLAMService(t, attrCfg, logger, false, testExecutableName)
+		test.That(t, err, test.ShouldBeNil)
+
+		timestamp1, _ = svc.GetLatestMapInfo(context.Background())
+		svc.Reconfigure(context.Background(), resource.Dependencies{}, resource.Config{})
```
jeremyrhyde, line 426: check error of reconfigure

jeremyrhyde (changes requested):
LGTM, a few small nits and error check improvements but no large issues left. PR is looking great!

viam-cartographer.go
PStrutz: Nope will do that

viam-cartographer_test.go
PStrutz: ok

viam-cartographer_test.go
PStrutz: I tried that, but because they are of type time.Time, the test package doesn't know how t compare them

viam-cartographer.go
```diff
@@ -142,6 +142,8 @@ func New(
 		cancelFunc:            cancelFunc,
 		logger:                logger,
 		bufferSLAMProcessLogs: bufferSLAMProcessLogs,
+		localizationMode:      mapRateSec == 0,
+		mapTimestamp:          time.Now(),
```
jeremyrhyde, line 146:
```suggestion
		mapTimestamp:          time.Now().UTC(),
```

viam-cartographer.go
jeremyrhyde, line 347:
```suggestion
	cartoSvc.mapTimestamp = time.Now().UTC()
```

PStrutz: As far as I can tell, the error only raises an error and does nothing else (i.e. no calls to New). It seems the expectation is that models implement their own versions of this error, and are advised they "should be able to reconfigure in place to support the best user experience" but this is not currently being done in cartographer.

jeremyrhyde (approved):
LGTM!

PR #160, opened 2023-06-23
[RSDK-3541] CartoFacade GetPosition

nicksanford:
[Ticket](https://viam.atlassian.net/browse/RSDK-3541)

1. Change CartoFacade::GetPosition() to return the actual pose from cartographer, which is updated when CartoFacade::AddSensorReading is called.
2. Remove unused fields from viam_carto_get_position_response (those which are needed for the GetPositionResponse protobuf type but which are filled in using the go spatial math library, which is not available to the cartographer C++ code. 
3. Remove component_reference from the `viam_carto_config` C type. This was done b/c it was previously assumed to be equal to the first element of the sensors list without adding any additional information. Removing it make sit clear that the component_reference is assumed to be the first element of the sensors list (which is the behavior of cartographer-module prior to full mod). 
4. Make it so that the CartoFacade.component_reference member is now a bstrting which is copied GetPosition responses are returned. This was done to not need to convert from bstring to std string only to then do another copy. This change also required adding a destructor to CartoFacade to clean up CartoFacade.component_reference  which is a C type that doesn't obey RAII.
5.  Add handling of nullptrs in C API

### DIFF
--- a/cartofacade/capi_test.go
+++ b/cartofacade/capi_test.go
@@ -29,8 +29,6 @@ func TestGetConfig(t *testing.T) {
 		dataDir := bstringToGoString(vcc.data_dir)
 		test.That(t, dataDir, test.ShouldResemble, dir)
 
-		componentReference := bstringToGoString(vcc.component_reference)
-		test.That(t, componentReference, test.ShouldResemble, "component")
 		freeBstringArray(vcc.sensors, vcc.sensors_len)
 
 		test.That(t, vcc.lidar_config, test.ShouldEqual, twoD)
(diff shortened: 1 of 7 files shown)


viam-cartographer/src/carto_facade/carto_facade.cc
```diff
@@ -63,11 +63,29 @@ std::ostream &operator<<(std::ostream &os, const ActionMode &action_mode) {
 std::string to_std_string(bstring b_str) {
     int len = blength(b_str);
     char *tmp = bstr2cstr(b_str, 0);
+    if (tmp == NULL) {
```
nicksanford, line 66: we need to return errors if bstring allocation fails

viam-cartographer/src/carto_facade/carto_facade.cc
```diff
@@ -84,9 +102,7 @@ config from_viam_carto_config(viam_carto_config vcc) {
     for (int i = 0; i < vcc.sensors_len; i++) {
         c.sensors.push_back(to_std_string(vcc.sensors[i]));
     }
-
     c.data_dir = to_std_string(vcc.data_dir);
-    c.component_reference = to_std_string(vcc.component_reference);
```
nicksanford, line 89: turns out that having a component_reference doesn't make sense given it is just always the first element in the sensors list

viam-cartographer/src/carto_facade/carto_facade.cc
```diff
@@ -98,6 +114,7 @@ config from_viam_carto_config(viam_carto_config vcc) {
     if (vcc.map_rate_sec < 0) {
         throw VIAM_CARTO_MAP_RATE_SEC_INVALID;
     }
+    c.component_reference = c.sensors[0];
```
nicksanford, line 117: We no longer take a component_reference from the vcc and instead infer it from the first element of the sensors list (as slam_service does today).

viam-cartographer/src/carto_facade/carto_facade_test.cc
```diff
@@ -11,25 +11,25 @@
...
 
+#include "bstrlib.h"
 #include "glog/logging.h"
 #include "test_helpers.h"
 
 namespace tt = boost::test_tools;
 namespace fs = std::filesystem;
 namespace bfs = boost::filesystem;
 namespace help = viam::carto_facade::test_helpers;
+const auto tol = tt::tolerance(0.001);
```
nicksanford, line 22: done to remove duplication

cartofacade/capi_test.go
```diff
@@ -161,23 +152,16 @@ func TestCGoAPI(t *testing.T) {
...
-		test.That(t, holder.Imag, test.ShouldEqual, 700)
-		test.That(t, holder.Jmag, test.ShouldEqual, 800)
-		test.That(t, holder.Kmag, test.ShouldEqual, 900)
-
-		test.That(t, holder.Theta, test.ShouldEqual, 1000)
-		test.That(t, holder.Real, test.ShouldEqual, 1100)
+		test.That(t, holder.Imag, test.ShouldEqual, 0)
+		test.That(t, holder.Jmag, test.ShouldEqual, 0)
+		test.That(t, holder.Kmag, test.ShouldEqual, 0)
+		test.That(t, holder.Real, test.ShouldEqual, 1)
```
kim-mishra, line 164: Why do these tests still work now that you are returning the real position?

viam-cartographer/src/carto_facade/carto_facade.cc
```diff
@@ -628,10 +637,9 @@ int CartoFacade::Stop() {
 };
 
 void CartoFacade::AddSensorReading(const viam_carto_sensor_reading *sr) {
-    std::string sensor = to_std_string(sr->sensor);
-    if (config.sensors[0] != sensor) {
-        VLOG(1) << "expected sensor: " << sensor << " to be "
-                << config.sensors[0];
+    if (biseq(config.component_reference, sr->sensor) == false) {
```
kim-mishra, line 640: What does this do?

nicksanford: this tests that two bstring s are not equal

cartofacade/capi_test.go
nicksanford: B/c it takes 3 lidar readings before GetPosition returns a non 0 result, see https://github.com/viamrobotics/viam-cartographer/pull/160/files#diff-0bd1cb9b5092e5c3c4e4ffa3be356ac99759fdc0c92546d1d7de11aed037dde0R602

kim-mishra: ohhhh I see, thanks!

kim-mishra (approved):
LGTM!

PR #172, opened 2023-06-28
RSDK-3828 - create cartofacade wrapper

kim-mishra:
Create wrapper around carto facade queue so that users have a clean API free of type casting


cartofacade/carto_facade_wrapper.go
```diff
@@ -0,0 +1,132 @@
+// Package cartofacade contains the api to call into CGO
```
kim-mishra, line 1: thoughts on file name?

cartofacade/capi.go
```diff
@@ -54,13 +54,13 @@ type CartoInterface interface {
...
-	getPosition() (GetPosition, error)
+	getPosition() (PositionInfo, error)
 	getPointCloudMap() ([]byte, error)
 	getInternalState() ([]byte, error)
 }
 
-// GetPosition holds values returned from c to be processed later
-type GetPosition struct {
+// PositionInfo holds values returned from c to be processed later
+type PositionInfo struct {
```
kim-mishra, line 63: Need to change this so that I can call the function in the API `GetPosition`, open to additional name suggestions

cartofacade/capi_test.go
```diff
@@ -138,13 +138,13 @@ func TestCGoAPI(t *testing.T) {
 		pc, err := pointcloud.ReadPCD(file)
 		test.That(t, err, test.ShouldBeNil)
 
-		// test invalid addSensorReading: valid reading binary
+		// test valid addSensorReading: valid reading binary
```
kim-mishra, line 141: Was this a typo? There are no errors from these calls

cartofacade/capi.go
nicksanford: I don't understand that. You should be able to have a function called GetPosition & a struct called GetPosition. If not, lets call it GetPositionResponse.

cartofacade/capi_test.go
nicksanford: yes it was a typo

cartofacade/carto_facade_wrapper.go
nicksanford: Why not just put all these in the existing https://github.com/viamrobotics/viam-cartographer/blob/main/cartofacade/carto_facade.go?

kim-mishra: I thought it was easier to read to have it separated out, so someone looking at this API wouldn't even need to see the queue implementation. Happy to change it if you think that would be better

cartofacade/capi.go
kim-mishra: I mustve misunderstood the error I was getting earlier, it is working now reverted

cartofacade/carto_facade_wrapper.go
nicksanford: I think it's confusing to have methods for a type defined in a different file than the type itself.
I think its ok to have both the public api & private methods in the same file as the fact that the private methods are private tell you as the API consumer you don't need to worry about them. I think putting the public methods at the top of the file also help with that.

I vote to put these methods in carto_facade.go

cartofacade/carto_facade_wrapper.go
```diff
@@ -0,0 +1,132 @@
...
+package cartofacade
+
+import (
+	"context"
+	"errors"
+	"sync"
+	"time"
+)
+
```
nicksanford: const emptyRequestParams := map[RequestParamType]interface{}{}

cartofacade/carto_facade_wrapper.go
```diff
@@ -0,0 +1,132 @@
...
+// Initialize calls into the cartofacade C code.
+func (cf *CartoFacade) Initialize(ctx context.Context, timeout time.Duration, activeBackgroundWorkers *sync.WaitGroup) error {
+	cf.start(ctx, activeBackgroundWorkers)
+
+	requestParams := map[RequestParamType]interface{}{}
+	untyped, err := cf.request(ctx, initialize, requestParams, timeout)
```
nicksanford:
```suggestion
	untyped, err := cf.request(ctx, initialize, emptyRequestParams, timeout)
```

cartofacade/carto_facade_wrapper.go
kim-mishra: done!

cartofacade/carto_facade_wrapper.go
```diff
@@ -0,0 +1,132 @@
...
+
+	cf.carto = &carto
+
+	return nil
+}
+
+// Start calls into the cartofacade C code.
+func (cf *CartoFacade) Start(ctx context.Context, timeout time.Duration) error {
+	requestParams := map[RequestParamType]interface{}{}
+	_, err := cf.request(ctx, start, requestParams, timeout)
```
nicksanford:
```suggestion
	_, err := cf.request(ctx, start, emptyRequestParams, timeout)
```

cartofacade/carto_facade_wrapper.go
```diff
@@ -0,0 +1,132 @@
...
+		return err
+	}
+
+	return nil
+}
+
+// Stop calls into the cartofacade C code.
+func (cf *CartoFacade) Stop(ctx context.Context, timeout time.Duration) error {
+	requestParams := map[RequestParamType]interface{}{}
+	_, err := cf.request(ctx, stop, requestParams, timeout)
```
nicksanford:
```suggestion
	_, err := cf.request(ctx, stop, emptyRequestParams, timeout)
```

cartofacade/carto_facade_wrapper.go
```diff
@@ -0,0 +1,132 @@
...
+		return err
+	}
+
+	return nil
+}
+
+// Terminate calls into the cartofacade C code.
+func (cf *CartoFacade) Terminate(ctx context.Context, timeout time.Duration) error {
+	requestParams := map[RequestParamType]interface{}{}
+	_, err := cf.request(ctx, terminate, requestParams, timeout)
```
nicksanford:
```suggestion
	_, err := cf.request(ctx, terminate, emptyRequestParams, timeout)
```

cartofacade/carto_facade_wrapper.go
```diff
@@ -0,0 +1,132 @@
...
+		return err
+	}
+
+	return nil
+}
+
+// GetPosition calls into the cartofacade C code.
+func (cf *CartoFacade) GetPosition(ctx context.Context, timeout time.Duration) (GetPosition, error) {
+	requestParams := map[RequestParamType]interface{}{}
+	untyped, err := cf.request(ctx, position, requestParams, timeout)
```
nicksanford:
```suggestion
	untyped, err := cf.request(ctx, position, emptyRequestParams, timeout)
```

cartofacade/carto_facade_wrapper.go
```diff
@@ -0,0 +1,132 @@
...
+		return GetPosition{}, errors.New("unable to cast response from cartofacade to a position info struct")
+	}
+
+	return pos, nil
+}
+
+// GetInternalState calls into the cartofacade C code.
+func (cf *CartoFacade) GetInternalState(ctx context.Context, timeout time.Duration) ([]byte, error) {
+	requestParams := map[RequestParamType]interface{}{}
+	untyped, err := cf.request(ctx, internalState, requestParams, timeout)
```
nicksanford:
```suggestion
	untyped, err := cf.request(ctx, internalState, emptyRequestParams, timeout)
```

cartofacade/carto_facade_wrapper.go
```diff
@@ -0,0 +1,132 @@
...
+		return []byte{}, errors.New("unable to cast response from cartofacade to a byte slice")
+	}
+
+	return internalState, nil
+}
+
+// GetPointCloudMap calls into the cartofacade C code.
+func (cf *CartoFacade) GetPointCloudMap(ctx context.Context, timeout time.Duration) ([]byte, error) {
+	requestParams := map[RequestParamType]interface{}{}
+	untyped, err := cf.request(ctx, pointCloudMap, requestParams, timeout)
```
nicksanford:
```suggestion
	untyped, err := cf.request(ctx, pointCloudMap, emptyRequestParams, timeout)
```

cartofacade/carto_facade.go
```diff
@@ -173,3 +173,126 @@ func (cf *CartoFacade) start(ctx context.Context, activeBackgroundWorkers *sync.
 		}
 	}()
 }
+
+// Initialize calls into the cartofacade C code.
+func (cf *CartoFacade) Initialize(ctx context.Context, timeout time.Duration, activeBackgroundWorkers *sync.WaitGroup) error {
+	cf.start(ctx, activeBackgroundWorkers)
+
+	requestParams := map[RequestParamType]interface{}{}
+	untyped, err := cf.request(ctx, initialize, requestParams, timeout)
```
nicksanford, line 182: How about we pull out all the empty requestParame into a package level constant?
```
const emptyRequestParams := map[RequestParamType]interface{}{}
```

```suggestion
	untyped, err := cf.request(ctx, initialize, emptyRequestParams, timeout)
```

cartofacade/carto_facade.go
```diff
@@ -173,3 +173,126 @@ func (cf *CartoFacade) start(ctx context.Context, activeBackgroundWorkers *sync.
 		}
 	}()
 }
+
+// Initialize calls into the cartofacade C code.
+func (cf *CartoFacade) Initialize(ctx context.Context, timeout time.Duration, activeBackgroundWorkers *sync.WaitGroup) error {
```
nicksanford, line 178: Can we please move all these public functions to the top of the file so that the package users can read all the public functions & types we expect them to use before needing to read any private ones?

cartofacade/carto_facade.go
```diff
@@ -173,3 +173,126 @@ func (cf *CartoFacade) start(ctx context.Context, activeBackgroundWorkers *sync.
...
+	cf.start(ctx, activeBackgroundWorkers)
```
nicksanford, line 179: Good idea doing this here

cartofacade/carto_facade.go
nicksanford, line 178: Question: Do we need to take the `activeBackgroundWorkers` as a parameter here or can we just create one within this function?

cartofacade/carto_facade_test.go
```diff
@@ -129,3 +133,396 @@ func TestRequest(t *testing.T) {
...
+	activeBackgroundWorkers := sync.WaitGroup{}
+
+	cfg, dir, err := GetTestConfig("mysensor")
+	algoCfg := GetTestAlgoConfig()
+	test.That(t, err, test.ShouldBeNil)
+	defer os.RemoveAll(dir)
+
+	cartoFacade := New(&lib, cfg, algoCfg)
+
+	t.Run("test Initialize - successful case", func(t *testing.T) {
```
nicksanford, line 151: Don't you also need to call terminate once you call initialize, otherwise you will leak memory at the C++ level?

nicksanford: Also, should this be mocking the Initialize C Function?

cartofacade/carto_facade.go
kim-mishra: I believe we want to take it in because we want to call `ActiveBackgroundWorkers.Wait()` outside of this function because if we initialize it here  and pass it in, and defer the call to  `ActiveBackgroundWorkers.Wait()`, it will block the function from returning

cartofacade/carto_facade_test.go
kim-mishra: Adding that in now!

kim-mishra: Is it possible to mock the C function, I wanted to but was not sure how to do it without the inject pattern

nicksanford: not really, lets talk about this

cartofacade/carto_facade_test.go
```diff
@@ -129,3 +133,404 @@ func TestRequest(t *testing.T) {
...
+	carto := CartoMock{}
+	carto.AddSensorReadingFunc = func(name string, reading []byte, time time.Time) error {
+		return nil
+	}
+	cartoFacade.carto = &carto
+	cartoFacade.Initialize(cancelCtx, 5*time.Second, &activeBackgroundWorkers)
+	defer cartoFacade.Terminate(cancelCtx, 5*time.Second)
+
+	t.Run("testing AddSensorReading", func(t *testing.T) {
+		timestamp := time.Date(2021, 8, 15, 14, 30, 45, 100, time.Local)
```
nicksanford, line 336: lets do this in UTC time rather than local time

kim-mishra: udpated, should we change it [here](https://github.com/viamrobotics/viam-cartographer/blob/e3977eb274ed9595aa68c6a7f4e0c1999f1fee71/cartofacade/capi_test.go#L65) and [here](https://github.com/viamrobotics/viam-cartographer/blob/e3977eb274ed9595aa68c6a7f4e0c1999f1fee71/cartofacade/capi_test.go#L119) or does it need to be local for sensor reading

nicksanford: yes please!

nicksanford (approved):
Looks great! Assuming the race conditions are resolved & tests are completing quickly, this is good to merge.

PR #177, opened 2023-06-28
RSDK-3545 - add sensor process

kim-mishra:
[Ticket](https://viam.atlassian.net/browse/RSDK-3545)

- `SensorProcess` polls the lidar to get the next sensor reading and adds it to the mapBuilder by calling `addSensorReading` in a loop and exits if the context provided is canceled. It also starts the background log process
- startBackgroundLogAggregator logs the failure counts on a time interval provided by the caller
- `addSensorReading` checks if the metadata from the GRPC call has a timestamp, if it does it calls `addSensorReadingFromReplaySensor` otherwise It calls `addSensorReadingFromLiveReadings`
- `addSensorReadingFromReplaySensor` continuously calls `cartofacade.AddSensorReading` until it can acquire lock, and increments the failure count of this time stamp each time it fails
- `addSensorReadingFromLiveReadings` calls `cartofacade.AddSensorReading` once, and skips the reading if it cannot acquire the lock, and increments the failure count of this time stamp once
- There is also a change to the testhelper file so that we can mock a replay sensor that returns a corrupted time stamp that would not be able to be parsed

Questions:
1. Right now, I keep track of the number of failures for each time stamp and on a frequency that can be set by the caller, log for each timestamp. Is this sufficient?


sensorprocess/sensorprocess.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+func incrementLogCount(params Params, t time.Time) {
+	_, ok := params.LogFailures[t]
+	if !ok {
+		params.LogFailures[t] = 1
+		return
+	}
+	params.LogFailures[t]++
+}
```
kim-mishra, line 148: I think when we log the failures,  it will not always be up to date because there is no lock right now.  I thought about adding a lock, but think it would cause performance issues and that given the aggregation of failures it is not as crucial to get the exact number of failures at that instant logged. Thoughts?

sensorprocess/sensorprocess.go
```diff
@@ -0,0 +1,154 @@
...
+func AddSensorReading(
+	params Params,
+) error {
+	reading, err := params.Lidar.GetData(params.Ctx)
+	if err != nil {
+		return err
+	}
+
+	buf := new(bytes.Buffer)
+	err = pointcloud.ToPCD(reading, buf, 0)
```
nicksanford, line 67:
```suggestion
	err = pointcloud.ToPCD(reading, buf, pointcloud.PCDBinary)
```

nicksanford:
PR moved to: https://github.com/viamrobotics/viam-cartographer/pull/185

PR #179, opened 2023-06-29
hotfix-fix-compile-error-on-mac

nicksanford:
I forgot to add a header include: https://github.com/viamrobotics/viam-cartographer/pull/160/files#diff-49d4a9777ef7773792665e3523fe1c5032de5373119d778dd3e461329a60399fR10 

This apparently is only caught by clang, which is only used on mac, which we don't test in CI.

Detected in this PR:
https://github.com/viamrobotics/viam-cartographer/pull/170/files#diff-49d4a9777ef7773792665e3523fe1c5032de5373119d778dd3e461329a60399fR4

This also changes the map_rate_sec from 1 to 60 to prevent flakey tests failures if it takes longer than a second to add all the lidar data which may cause add_lidar_reading to fail unexpectedly as the lock is held by the background thread which saves internal state to disk: https://github.com/viamrobotics/viam-cartographer/actions/runs/5414901563/jobs/9842617172?pr=179


jeremyrhyde (approved):
LGTM

PR #181, opened 2023-06-30
RSDK-3786 - use-cartofacade-for-init-and-term

kim-mishra:
[Ticket](https://viam.atlassian.net/browse/RSDK-3786)

Changes to `New`:
- If `modularizationV2Enabled == true`
    - initialize cartofacade library
    - initialize cartofacade
    - start background goroutine that enforces a single call into C at a time

Changes to `Close`:
- If `modularizationV2Enabled == true`
    - stop background goroutine (this will happen when `cartoSvc.cancelFunc()` is called)
    - terminate cartofacade
    - terminate cartofacade library

### DIFF
--- a/viam-cartographer.go
+++ b/viam-cartographer.go
@@ -171,12 +171,12 @@ func New(
 			Sensors:            svcConfig.Sensors,
 			MapRateSecond:      dataRateMsec * 1000,
 			DataDir:            svcConfig.DataDirectory,
-			ComponentReference: svcConfig.Sensors[0], // Can this be removed?
-			LidarConfig:        cartofacade.TwoD,     // Should this come from somewhere or is hardcoded ok for now?
+			ComponentReference: svcConfig.Sensors[0],
+			LidarConfig:        cartofacade.TwoD,
 		}
 
 		// Get the algo config
-		cartoAlgoCfg := cartofacade.CartoAlgoConfig{} // Where are algo config params being set
+		cartoAlgoCfg := cartofacade.CartoAlgoConfig{}
 
 		// Initialize cartofacade
 		cf := cartofacade.New(&cartofacade.Lib, cartoCfg, cartoAlgoCfg)
--- a/viam-cartographer.go
+++ b/viam-cartographer.go
@@ -44,7 +44,7 @@ const (
 	defaultSensorValidationIntervalSec   = 1
 	parsePortMaxTimeoutSec               = 60
 	localhost0                           = "localhost:0"
-	Timeout                              = 5 * time.Second // I think this is the last place we can bubble this up, thoughts on what the actual value should be?
+	Timeout                              = 5 * time.Second
 )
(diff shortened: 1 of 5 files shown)


viam-cartographer.go
```diff
@@ -43,6 +44,7 @@ const (
 	defaultSensorValidationIntervalSec   = 1
 	parsePortMaxTimeoutSec               = 60
 	localhost0                           = "localhost:0"
+	Timeout                              = 5 * time.Second // I think this is the last place we can bubble this up, thoughts on what the actual value should be?
```
kim-mishra, line 47: I think this is the last place we can bubble this up, thoughts on what the actual value should be?

viam-cartographer.go
```diff
@@ -43,6 +44,7 @@ const (
 	defaultSensorValidationIntervalSec   = 1
 	parsePortMaxTimeoutSec               = 60
 	localhost0                           = "localhost:0"
+	Timeout                              = 5 * time.Second
```
kim-mishra, line 47: I think this is the last place we can bubble this up, thoughts on what the actual value should be?

viam-cartographer.go
```diff
@@ -153,28 +156,57 @@ func New(
...
 
-	if err := cartoSvc.StartSLAMProcess(ctx); err != nil {
-		return nil, errors.Wrap(err, "error with slam service slam process")
-	}
+		// Get the config
+		cartoCfg := cartofacade.CartoConfig{
+			Sensors:            svcConfig.Sensors,
+			MapRateSecond:      dataRateMsec * 1000,
+			DataDir:            svcConfig.DataDirectory,
+			ComponentReference: svcConfig.Sensors[0],
```
kim-mishra, line 174: @nicksanford can `ComponentReference` be removed now with your changes to use `sensors[0]` in place of every where it was used, or is it needed for something else?

viam-cartographer.go
```diff
@@ -153,28 +156,57 @@ func New(
...
+			LidarConfig:        cartofacade.TwoD,
```
kim-mishra, line 175: Is it okay that this is hardcoded or should it come from info in the config?

viam-cartographer.go
```diff
@@ -153,28 +156,57 @@ func New(
...
+		}
 
-	client, clientClose, err := vcConfig.SetupGRPCConnection(ctx, cartoSvc.port, dialMaxTimeoutSec, logger)
-	if err != nil {
-		return nil, errors.Wrap(err, "error with initial grpc client to slam algorithm")
-	}
-	cartoSvc.clientAlgo = client
-	cartoSvc.clientAlgoClose = clientClose
+		// Get the algo config
+		cartoAlgoCfg := cartofacade.CartoAlgoConfig{}
```
kim-mishra, line 179: Where are the params for algo config params coming from?

PR #184, opened 2023-07-05
[RSDK-3427] confirm-carto-facade-is-in-correct-state

nicksanford:
[Ticket](https://viam.atlassian.net/browse/RSDK-3427)

1. Change the CartoFacade to represent its state using an enum rather than a `started` boolean
2. Throw in each method called by the C API if the CartoFacade object is not in the expected state for that method


viam-cartographer/src/carto_facade/carto_facade.cc
```diff
@@ -82,6 +82,21 @@ std::ostream &operator<<(std::ostream &os, const ActionMode &action_mode) {
     return os;
 }
 
+std::ostream &operator<<(std::ostream &os, const CartoFacadeState &state) {
```
nicksanford, line 85: Added so we can log the CartoFacadeState

viam-cartographer/src/carto_facade/carto_facade.h
```diff
@@ -359,7 +356,7 @@ class CartoFacade {
     viam::carto_facade::config config;
     viam_carto_algo_config algo_config;
     std::string path_to_internal_state;
-    std::atomic<bool> started{false};
```
nicksanford, line 362: I changed this atomic boolean to be an atomic enum to be able to better describe the different possible states of the CartoFacade.

viam-cartographer/src/carto_facade/carto_facade_test.cc
```diff
@@ -444,7 +444,7 @@ BOOST_AUTO_TEST_CASE(CartoFacade_init_terminate) {
     auto path_to_internal_state = tmp_dir / fs::path("internal_state");
 
     BOOST_TEST((cf->path_to_internal_state) == path_to_internal_state.string());
-    BOOST_TEST((cf->started) == false);
+    BOOST_TEST(((cf->state) == CartoFacadeState::IO_INITIALIZED));
```
nicksanford, line 447: Fun fact, you need to provide an outer paren when using boost test with enum classes: https://stackoverflow.com/questions/30860241/use-enum-classes-with-boost-test

viam-cartographer/src/carto_facade/carto_facade.cc
```diff
@@ -289,7 +309,7 @@ void CartoFacade::IOInit() {
     }
 
     // TODO: google cartographer will termiante the program if
-    // the internal state is invalid
```
EshaMaharishi, line 292: Hmm, are you sure this should say carto facade? I thought this was talking about google cartographer's internal state

viam-cartographer/src/carto_facade/carto_facade.cc
```diff
@@ -689,16 +729,30 @@ void CartoFacade::SaveInternalStateOnInterval() {
...
+        if (state != CartoFacadeState::STARTED) {
             LOG(INFO) << "Finished saving final optimized internal state";
             break;
         }
     }
 }
 
-void CartoFacade::Stop() { StopSaveInternalState(); };
+void CartoFacade::Stop() {
+    if (state != CartoFacadeState::STARTED) {
```
EshaMaharishi, line 740: Can we call Stop when the state is INITIALIZED or IO_INITIALIZED?

viam-cartographer/src/carto_facade/carto_facade.cc
```diff
@@ -689,16 +729,30 @@ void CartoFacade::SaveInternalStateOnInterval() {
...
+        LOG(ERROR) << "carto facade is in state: " << state << " expected "
+                   << CartoFacadeState::STARTED;
+        throw VIAM_CARTO_NOT_IN_STARTED_STATE;
+    }
+    state = CartoFacadeState::IO_INITIALIZED;
```
EshaMaharishi, line 745: Should we set it back to INITIALIZED?

EshaMaharishi (approved):
LGTM with optional comments

viam-cartographer/src/carto_facade/carto_facade.cc
nicksanford: Nope. Stop() stops the background thread which persists internal state to disk. Calling Stop() in any state other than `STARTED` means that there will be no background thread to stop.

viam-cartographer/src/carto_facade/carto_facade.cc
nicksanford: fixed

viam-cartographer/src/carto_facade/carto_facade.cc
EshaMaharishi: That makes sense, since we're not destroying CartoFacade here, just stopping the background thread that gets started in Start. Thanks for explaining!

EshaMaharishi (approved):

PR #191, opened 2023-07-11
add jobs done file

JohnN193:
adds a jobs done file when the replay camera stops. additionally prevents cartographer from repeating the warning about the job being completed.

Additional issue is this file gets created when all lidar scans are read into the job, but this does not mean that all lidar scans have been processed

unsure of interactions with full mod and interactions with a live replay camera

### DIFF
--- a/viam-cartographer.go
+++ b/viam-cartographer.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"io"
 	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"sync"
@@ -38,6 +39,7 @@ import (
 var (
 	Model    = resource.NewModel("viam", "slam", "cartographer")
 	cartoLib cartofacade.CartoLib
+	jobsDone = false
 )
 
 const (
@@ -637,7 +639,19 @@ func (cartoSvc *cartographerService) readDataOnInterval(ctx context.Context, lid
 
 func (cartoSvc *cartographerService) getNextDataPoint(ctx context.Context, lidar lidar.Lidar, c chan int) {
 	if _, err := dim2d.GetAndSaveData(ctx, cartoSvc.dataDirectory, lidar, cartoSvc.logger); err != nil {
-		cartoSvc.logger.Warn(err)
+		if strings.Contains(err.Error(), "reached end of dataset") {
+			if !jobsDone {
+				cartoSvc.logger.Warn(err)
+				_, err = os.Create(filepath.Join(cartoSvc.dataDirectory, "Jobs_Done.txt"))
+				if err != nil {
+					cartoSvc.logger.Warn(err)
+				}
+				jobsDone = true
+			}
+		} else {
+			cartoSvc.logger.Warn(err)
+		}
+
 	}
 	if c != nil {
 		c <- 1
--- a/viam-cartographer.go
+++ b/viam-cartographer.go
@@ -651,7 +651,6 @@ func (cartoSvc *cartographerService) getNextDataPoint(ctx context.Context, lidar
 		} else {
 			cartoSvc.logger.Warn(err)
 		}
-
 	}
 	if c != nil {
 		c <- 1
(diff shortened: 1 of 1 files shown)


viam-cartographer.go
```diff
@@ -637,7 +639,18 @@ func (cartoSvc *cartographerService) readDataOnInterval(ctx context.Context, lid
...
 	if _, err := dim2d.GetAndSaveData(ctx, cartoSvc.dataDirectory, lidar, cartoSvc.logger); err != nil {
-		cartoSvc.logger.Warn(err)
+		if strings.Contains(err.Error(), "reached end of dataset") {
+			if !jobsDone {
+				cartoSvc.logger.Warn(err)
+				_, err = os.Create(filepath.Join(cartoSvc.dataDirectory, "Jobs_Done.txt"))
+				if err != nil {
+					cartoSvc.logger.Warn(err)
+				}
+				jobsDone = true
```
nicksanford, line 649: 1. Could we talk about the requirements for this feature so I can be sure that it gets rolled into full mod?
2. I don't think it works to track a given session completing at the package level, as variable is shared across viam cartographer instances & persists even after reconfiguration. 
3. I recommend instead (assuming this should be scoped to a given viam cartographer instance & should be reset on restart) to put it on thecartographerService struct. 
4. Also, should Jobs_Done.txt still persist if cartographer is reconfigured?
cc @EshaMaharishi

EshaMaharishi, line 649: [note] This effectively makes [the loop that calls this function](https://github.com/viamrobotics/viam-cartographer/blob/8c0ff9f173bb3f49e57c9751d19b03dfa4277f46/viam-cartographer.go#L595-L604) spin continuously without doing any work, which will waste CPU.

However, it's not so different from what was happening before this patch (now we are simply not logging in each iteration of the loop). Also, we should only see "reached end of dataset" in offline mode, which will only run in the cloud and the job will exit very soon after we see that error.

So, I think it's okay to leave this as-is.

viam-cartographer.go
```diff
@@ -637,7 +639,18 @@ func (cartoSvc *cartographerService) readDataOnInterval(ctx context.Context, lid
 
 func (cartoSvc *cartographerService) getNextDataPoint(ctx context.Context, lidar lidar.Lidar, c chan int) {
 	if _, err := dim2d.GetAndSaveData(ctx, cartoSvc.dataDirectory, lidar, cartoSvc.logger); err != nil {
-		cartoSvc.logger.Warn(err)
+		if strings.Contains(err.Error(), "reached end of dataset") {
+			if !jobsDone {
+				cartoSvc.logger.Warn(err)
+				_, err = os.Create(filepath.Join(cartoSvc.dataDirectory, "Jobs_Done.txt"))
```
EshaMaharishi, line 645: [nit] Jobs_Done.txt -> job_done.txt (no 's' and lower case)

EshaMaharishi (approved):
Lgtm for the demo, but would you mind filing a ticket to move writing the job_done file to after the final lidar scan has been processed?

viam-cartographer.go
EshaMaharishi: It doesn't matter if the jobs done file persists after cartographer is reconfigured since a cloud slam session will only be configured once, but I agree for code cleanliness it's better to scope it to a single configuration

JohnN193: yeah I did not want to dive deep to get cartographer to stop the data thread, especially with the move to full mod coming up

viam-cartographer.go
JohnN193: kk I'll swap it, tbh I have been saying job's done because its a [reference to a game](https://www.youtube.com/watch?v=bBCB0y9Mgns)

JohnN193:
https://viam.atlassian.net/browse/RSDK-3980 ticket to move the job creation file

nroede (approved):
LGTM

viam-cartographer.go
JohnN193: followed up offline, in the followup ticket we will switch to using a DoCommand

PR #193, opened 2023-07-12
RSDK-3965 - call into cgo for getPointCloudMap

kim-mishra:
[RSDK-3695](https://viam.atlassian.net/browse/RSDK-3965)
- If the modularizationV2Enabled feature flag is true, call into the new cartofacade API for GetPointCloudMap
- Write unit tests to make sure cgo is returning the same output that the original code would have

[RSDK-3695]: https://viam.atlassian.net/browse/RSDK-3695?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/viam-cartographer.go
+++ b/viam-cartographer.go
@@ -4,6 +4,7 @@ package viamcartographer
 
 import (
 	"bufio"
+	"bytes"
 	"context"
 	"io"
 	"os"
@@ -36,8 +37,9 @@ import (
 
 // Model is the model name of cartographer.
 var (
-	Model    = resource.NewModel("viam", "slam", "cartographer")
-	cartoLib cartofacade.CartoLib
+	Model          = resource.NewModel("viam", "slam", "cartographer")
+	cartoLib       cartofacade.CartoLib
+	chunkSizeBytes = 1 * 1024 * 1024
 )
 
 const (
@@ -539,10 +541,33 @@ func (cartoSvc *cartographerService) GetPointCloudMap(ctx context.Context) (func
 	ctx, span := trace.StartSpan(ctx, "viamcartographer::cartographerService::GetPointCloudMap")
 	defer span.End()
 
+	if cartoSvc.modularizationV2Enabled {
+		return cartoSvc.getPointCloudMapModularizationV2(ctx)
+	}
+
 	if !cartoSvc.localizationMode {
 		cartoSvc.mapTimestamp = time.Now().UTC()
 	}
 	return grpchelper.GetPointCloudMapCallback(ctx, cartoSvc.Name().ShortName(), cartoSvc.clientAlgo)
+}
+
+func (cartoSvc *cartographerService) getPointCloudMapModularizationV2(ctx context.Context) (func() ([]byte, error), error) {
+	chunk := make([]byte, chunkSizeBytes)
+	pc, err := cartoSvc.cartofacade.GetPointCloudMap(ctx, cartoSvc.cartoFacadeTimeout)
+	if err != nil {
+		return nil, err
+	}
+
+	pointcloudReader := bytes.NewReader(pc)
+
+	f := func() ([]byte, error) {
+		bytesRead, err := pointcloudReader.Read(chunk)
+		if err != nil {
+			return nil, err
+		}
+		return chunk[:bytesRead], err
+	}
+	return f, nil
 }
 
 // GetInternalState creates a request, calls the slam algorithms GetInternalState endpoint and returns a callback
--- a/viam-cartographer.go
+++ b/viam-cartographer.go
@@ -549,13 +549,14 @@ func (cartoSvc *cartographerService) GetPointCloudMap(ctx context.Context) (func
 	ctx, span := trace.StartSpan(ctx, "viamcartographer::cartographerService::GetPointCloudMap")
 	defer span.End()
 
-  if cartoSvc.closed {
+	if cartoSvc.closed {
 		cartoSvc.logger.Warn("GetPointCloudMap called after closed")
 		return nil, ErrClosed
 	}
-  
+
 	if cartoSvc.modularizationV2Enabled {
 		return cartoSvc.getPointCloudMapModularizationV2(ctx)
+	}
 
 	if !cartoSvc.localizationMode {
 		cartoSvc.mapTimestamp = time.Now().UTC()
--- a/viam-cartographer.go
+++ b/viam-cartographer.go
@@ -38,9 +38,8 @@ import (
 
 // Model is the model name of cartographer.
 var (
-	Model          = resource.NewModel("viam", "slam", "cartographer")
-	cartoLib       cartofacade.CartoLib
-	chunkSizeBytes = 1 * 1024 * 1024
+	Model    = resource.NewModel("viam", "slam", "cartographer")
+	cartoLib cartofacade.CartoLib
 	// ErrClosed denotes that the slam service method was called on a closed slam resource.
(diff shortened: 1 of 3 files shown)


viam-cartographer_test.go
```diff
@@ -351,8 +351,8 @@ func TestNew(t *testing.T) {
 		// TODO: Implement these
 		// timestamp1, err := svc.GetLatestMapInfo(context.Background())
 		// test.That(t, err, test.ShouldBeNil)
-		// _, err = svc.GetPointCloudMap(context.Background())
-		// test.That(t, err, test.ShouldBeNil)
+		_, err = svc.GetPointCloudMap(context.Background())
```
nicksanford, line 354: Can you please also add the position?

viam-cartographer.go
```diff
@@ -555,12 +557,35 @@ func (cartoSvc *cartographerService) GetPointCloudMap(ctx context.Context) (func
 		return nil, ErrClosed
 	}
 
+	if cartoSvc.modularizationV2Enabled {
```
nicksanford, line 560: this needs to be after the `if !cartoSvc.localizationMode {` if statement

viam-cartographer_test.go
nicksanford: why are the tests on line 383 not implemented?

viam-cartographer.go
```diff
@@ -36,8 +37,9 @@ import (
 
 // Model is the model name of cartographer.
 var (
-	Model    = resource.NewModel("viam", "slam", "cartographer")
-	cartoLib cartofacade.CartoLib
+	Model          = resource.NewModel("viam", "slam", "cartographer")
+	cartoLib       cartofacade.CartoLib
+	chunkSizeBytes = 1 * 1024 * 1024
```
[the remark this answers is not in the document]
nicksanford: I recommend we just make this constant & not test by changing the chunk size

viam-cartographer_internal_test.go
```diff
@@ -428,6 +428,81 @@ func TestGetPointCloudMapEndpoint(t *testing.T) {
...
+	svc := &cartographerService{Named: resource.NewName(slam.API, "test").AsNamed()}
+	mockCartoFacade := &cartofacade.Mock{}
+
+	svc.cartofacade = mockCartoFacade
+	svc.modularizationV2Enabled = true
+
+	inputPointCloudMapBytes := []byte{1, 2, 3, 4, 5, 6, 7, 8, 9, 10}
+	setMockGetPointCloudFunc(mockCartoFacade, inputPointCloudMapBytes)
+
+	t.Run("small chunk size success", func(t *testing.T) {
```
nicksanford, line 453: Can we test the behavior when the response size is above & below the chunk limit with real-world files, i.e. without needing to modify the chunk size?  

Something like:
 `.artifact/data/viam-cartographer/outputs/viam-office-02-22-3/pointcloud/pointcloud_0.pcd` for a file below the 1 MB limit & 
`.artifact/data/viam-cartographer/outputs/viam-office-02-22-3/pointcloud/pointcloud_1.pcd` for a file over the limit.

viam-cartographer_internal_test.go
```diff
@@ -428,6 +428,81 @@ func TestGetPointCloudMapEndpoint(t *testing.T) {
...
+	t.Run("large chunk size success", func(t *testing.T) {
+		chunkSizeBytes = 100
+		callback, err := svc.GetPointCloudMap(context.Background())
+		test.That(t, err, test.ShouldBeNil)
+		pointCloudMapBytes, err := slam.HelperConcatenateChunksToFull(callback)
+		test.That(t, err, test.ShouldBeNil)
+		test.That(t, pointCloudMapBytes, test.ShouldResemble, inputPointCloudMapBytes)
+	})
+
+	t.Run("no bytes success", func(t *testing.T) {
```
nicksanford, line 471: in what way is this returning no bytes?

viam-cartographer.go
```diff
@@ -555,12 +557,35 @@ func (cartoSvc *cartographerService) GetPointCloudMap(ctx context.Context) (func
...
 		cartoSvc.mapTimestamp = time.Now().UTC()
 	}
 	return grpchelper.GetPointCloudMapCallback(ctx, cartoSvc.Name().ShortName(), cartoSvc.clientAlgo)
 }
 
+func (cartoSvc *cartographerService) getPointCloudMapModularizationV2(ctx context.Context) (func() ([]byte, error), error) {
+	chunk := make([]byte, chunkSizeBytes)
+	pc, err := cartoSvc.cartofacade.GetPointCloudMap(ctx, cartoSvc.cartoFacadeTimeout)
+	if err != nil {
+		return nil, err
```
nicksanford, line 574: where is the test for this caes?

viam-cartographer_internal_test.go
nicksanford: also, please have the test confirm that those files are indeed below & above the chunk size limit, respectively.

viam-cartographer_internal_test.go
kim-mishra: Where we set `inputPointCloudMapBytes` to empty and update the mock func to return that

```
inputPointCloudMapBytes = []byte{}

mockCartoFacade.GetPointCloudMapFunc = func(
	ctx context.Context,
	timeout time.Duration,
) ([]byte, error) {
	return inputPointCloudMapBytes, nil
}
```

viam-cartographer.go
kim-mishra: [here](https://github.com/viamrobotics/viam-cartographer/blob/f5ae0a00f9487e2e41268baa2106de10e15ba7cd/viam-cartographer_internal_test.go#L489) I mock `cartoSvc.cartofacade.GetPointCloudMap` to return an error. Is this what you mean?

viam-cartographer.go
```diff
@@ -54,6 +55,7 @@ const (
 	parsePortMaxTimeoutSec               = 60
 	localhost0                           = "localhost:0"
 	defaultCartoFacadeTimeout            = 5 * time.Second
+	chunkSizeBytes                       = 1 * 1024 * 1024
```
nicksanford, line 58: this should be const

kim-mishra: isnt it a const if it is in the const block?

nicksanford: 🤦 yes, yes it is... ignore!

viam-cartographer_internal_test.go
```diff
@@ -428,6 +430,83 @@ func TestGetPointCloudMapEndpoint(t *testing.T) {
...
+}
+
+func TestGetPointCloudMapEndpointModularizationV2Endpoint(t *testing.T) {
+	svc := &cartographerService{Named: resource.NewName(slam.API, "test").AsNamed()}
+	mockCartoFacade := &cartofacade.Mock{}
+
+	svc.cartofacade = mockCartoFacade
+	svc.modularizationV2Enabled = true
+
+	inputPointCloudMapBytes := []byte{}
```
nicksanford, line 452: I think you can delete this variable. Why do you need to share this empty byte slice between tests?

In general I try to have all test setup variables which are used by tests always be immutable, to prevent one test from affecting another test.

kim-mishra: leftover from trying to match old tests - removed!

nicksanford (approved):

PR #194, opened 2023-07-12
[not for merging] POC for IMU integration into SLAM

PStrutz:


config/config.go
```diff
@@ -86,20 +88,24 @@ func (config *Config) Validate(path string) ([]string, error) {
...
+	if config.IMUDataRateMsec < 0 {
+		return nil, errors.New("cannot specify imu_data_rate_msec less than zero")
+	}
+
 	if config.MapRateSec != nil && *config.MapRateSec < 0 {
 		return nil, errors.New("cannot specify map_rate_sec less than zero")
 	}
 
-	deps := config.Sensors
+	deps := config.LidarSensors
```
jeremyrhyde, line 99: need to add imu dependencies

dataprocess/dataprocess.go
```diff
@@ -31,6 +34,23 @@ func WritePCDToFile(pointcloud pc.PointCloud, filename string) error {
 	return WriteBytesToFile(buf.Bytes(), filename)
 }
 
+// WriteJSONToFile encodes the imu data and then saves it to the passed filename.
+func WriteJSONToFile(linearAcceleration r3.Vector, angularVelocity spatialmath.AngularVelocity, filename string) error {
+	buf := new(bytes.Buffer)
+
+	jsonLinAcc, err := json.Marshal(linearAcceleration)
```
jeremyrhyde, line 41: one thing to check is if it is better to created a combined json as opposed to writing twice. This can be done by creating an imuData object which is a `map[string]r3.Vector`.

Not necessary just could help with parsing if you start experiencing issues

sensors/imu/imu.go
```diff
@@ -0,0 +1,104 @@
+// Package imu implements the IMU sensor.
+package imu
```
jeremyrhyde, line 2: looks good!

viam-cartographer.go
```diff
@@ -317,27 +343,63 @@ func (cartoSvc *cartographerService) readDataOnInterval(ctx context.Context, lid
...
+				cartoSvc.getNextDataPoint(ctx, sensors, c)
 			})
 		}
 	}
 }
 
-func (cartoSvc *cartographerService) getNextDataPoint(ctx context.Context, lidar lidar.Lidar, c chan int) {
-	if _, err := dim2d.GetAndSaveData(ctx, cartoSvc.dataDirectory, lidar, cartoSvc.logger); err != nil {
-		cartoSvc.logger.Warn(err)
+func (cartoSvc *cartographerService) getNextDataPoint(ctx context.Context, sensors *slamSensors, c chan int) {
```
jeremyrhyde, line 389: [this should work! I'd personally have split off into `getNextLidarDataPoint` and `getNextImuDataPoint` so you dont need to check lidar_ative and imu_active. 

When I first saw this I thought this was a way to 'turn' off one of the sensors for testing.

But again its a POC and should work as is if you feel strongly about it

jeremyrhyde: A cool alternative is also to sent the `dim2d.GetAndSaveLidarData` and `imu.GetAndSaveIMUData` fucntions as inputs to GetNextDataPoint and to have those functions grab the required sensor (sensors.lidar or sensor.imu).

You would need to change the function to take in your slamSensors instead of lidar.Lidar and imu.Imu

jeremyrhyde (commented):
Initial comments on DataProcess

PR #195, opened 2023-07-12
RSDK-3992 - call into cgo for GetInternalState

kim-mishra:
[Ticket](https://viam.atlassian.net/browse/RSDK-3992)
- If the modularizationV2Enabled feature flag is true, call into the new cartofacade API for GetInternalState
- Write unit tests to make sure cgo is returning the same output that the original code would have

### DIFF
--- a/viam-cartographer_internal_test.go
+++ b/viam-cartographer_internal_test.go
@@ -460,14 +460,14 @@ func testApisThatReturnCallbackFuncsSuccess(
 	}
 
 	setMockFunc(mockCartoFacade, bytes)
-
-	getAndCheckCallbackFunc(t, api, false)
+	getAndCheckCallbackFunc(t, api, false, bytes)
 }
 
 func getAndCheckCallbackFunc(
 	t *testing.T,
 	api func(context.Context) (func() ([]byte, error), error),
 	emptyBytes bool,
+	inputBytes []byte,
 ) {
 	callback, err := api(context.Background())
 	test.That(t, err, test.ShouldBeNil)
@@ -478,6 +478,7 @@ func getAndCheckCallbackFunc(
 		test.That(t, bytes, test.ShouldBeNil)
 	} else {
 		test.That(t, bytes, test.ShouldNotBeNil)
+		test.That(t, bytes, test.ShouldResemble, inputBytes)
 	}
 }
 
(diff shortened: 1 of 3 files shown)


viam-cartographer_internal_test.go
```diff
@@ -428,6 +430,130 @@ func TestGetPointCloudMapEndpoint(t *testing.T) {
...
+			api,
+			true,
+		)
+	})
+
+	t.Run("no bytes success", func(t *testing.T) {
+		setMockFunc(mockCartoFacade, []byte{})
+		getAndCheckCallbackFunc(t, api, true)
+	})
+}
```
kim-mishra, line 524: Some of the logic here is a little strange, but I needed to do it to avoid getting a lint issue with duplication. LMK what you think

viam-cartographer_internal_test.go
```diff
@@ -442,50 +442,101 @@ func setMockGetPointCloudFunc(
...
-		inputPointCloudMapBytes, err := os.ReadFile(artifact.MustPath(file))
-		test.That(t, err, test.ShouldBeNil)
-		test.That(t, len(inputPointCloudMapBytes), test.ShouldBeGreaterThan, 1024*1024)
-
-		setMockGetPointCloudFunc(mockCartoFacade, inputPointCloudMapBytes)
-		callback, err := svc.GetPointCloudMap(context.Background())
-		test.That(t, err, test.ShouldBeNil)
-		pointCloudMapBytes, err := slam.HelperConcatenateChunksToFull(callback)
-		test.That(t, err, test.ShouldBeNil)
-		test.That(t, pointCloudMapBytes, test.ShouldResemble, inputPointCloudMapBytes)
```
nicksanford, line 477: Where are we actually asserting that both the output of the GetPointCloudMap & GetInternalState call back functions, when concatenated together, both above & below the chunk limit, equal the files they were derived from?

kim-mishra: i must've accidentally removed that during the refactor fixed and happening [here](https://github.com/viamrobotics/viam-cartographer/blob/88baa8b90fea27333d91f60991b0d572bc26d522/viam-cartographer_internal_test.go#L481)

nicksanford (approved):

PR #186, opened 2023-07-07
[RSDK-3786] full-mod-new-close

nicksanford:
[Ticket](https://viam.atlassian.net/browse/RSDK-3786)

1. Make `viamcartographer.New`  initialize 1.) a carto facade instance 2.) a background go routine which will call into the carto facade C code 3.) a sensor process go routine to add sensor data when the full mod feature flag is enabled
2. Make `viamcartographer.Close` terminate in the proper order the resources initialized by `viamcartographer.New`
3. Add algo config defaults & parsing logic (pre full mod this was a C++ code responsibility, now it is a go code responsibility)
4. Don't setup deprecated directory structure if full mod feature flag is enabled, as otherwise it always causes full mod carto to throw errors on initialization (as per the project requirements)
5. Change package structure to fix circular dependency issues which would otherwise be caused by making viam-cartographer dependent on the sensorprocess package.
6. Change Makefile's `build` target to compile the C++ code (if not already compiled) before compiling the go code. This is needed as now main.go needs to link in C code, which was not the case prior. W/o this change, it doesn't build.

PR #203, opened 2023-07-17
fix-makefile-not-rebuilding-cartographer-module

nicksanford:
Fixes `make build` not rebuilding cartographer-module if it has been built before. 
Now `make build` always rebuilds the cartographer-module go binary.
Follows the same pattern as: viam-server's `make server` make target: https://github.com/viamrobotics/rdk/blob/main/Makefile#L81
I accidentally introduced this bug in the build system in https://github.com/viamrobotics/viam-cartographer/pull/186

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -95,12 +95,13 @@ else
 	$(error "Unsupported system. Only apt and brew currently supported.")
 endif
 
-build: bin/cartographer-module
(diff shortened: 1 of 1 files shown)


nicksanford:
moved here: https://github.com/viamrobotics/viam-cartographer/pull/205

PR #206, opened 2023-07-18
RSDK-3980 - switch job done file creation to do command

kim-mishra:
# Changes
- If the error with "reached end of dataset" is seen while reading data from the lidar, stop the sensor process
- If sending job_done command to DoCommand, be able to tell if the job has finished or not

# Testing

Following log only appears once
```
\_ 2023-07-18T15:38:28.571-0400	WARN	cartographerModule	sensorprocess/sensorprocess.go:54	NextPointCloud error: rpc error: code = Unknown desc = reached end of dataset
```

Results from grpcurl to DoCommand:
```
// Before above log
➜  ~ viam --base-url=https://pr-2017-appmain-bplesliplq-uc.a.run.app:443 robot part run --robot kims-robot --location "First Location" --organization "Nils's Org" -part kims-robot-main  -d '{"name": "slam-svc", "command": {"job_done": "true"}}' viam.service.slam.v1.SLAMService.DoCommand
{
  "result": {
      "job_done": false
    }
}
```

```
// After above log
➜  ~ viam --base-url=https://pr-2017-appmain-bplesliplq-uc.a.run.app:443 robot part run --robot kims-robot --location "First Location" --organization "Nils's Org" -part kims-robot-main  -d '{"name": "slam-svc", "command": {"job_done": "true"}}' viam.service.slam.v1.SLAMService.DoCommand
{
  "result": {
      "job_done": true
    }
}
```

### DIFF
--- a/sensorprocess/sensorprocess.go
+++ b/sensorprocess/sensorprocess.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"errors"
 	"math"
+	"strings"
 	"time"
 
 	"github.com/edaniels/golog"
@@ -22,6 +23,7 @@ type Config struct {
 	Timeout          time.Duration
 	Logger           golog.Logger
 	TelemetryEnabled bool
+	jobDone          bool
 }
 
 // Start polls the lidar to get the next sensor reading and adds it to the cartofacade.
@@ -29,11 +31,11 @@ type Config struct {
 func Start(
 	ctx context.Context,
 	config Config,
-) {
+) bool {
 	for {
 		select {
 		case <-ctx.Done():
-			return
+			return config.jobDone
 		default:
 			addSensorReading(ctx, config)
 		}
@@ -47,7 +49,14 @@ func addSensorReading(
 ) {
 	tsr, err := config.Lidar.TimedSensorReading(ctx)
 	if err != nil {
-		config.Logger.Warn(err)
+		if strings.Contains(err.Error(), "reached end of dataset") {
+			if !config.jobDone {
+				config.Logger.Warn(err)
+				config.jobDone = true
+			}
+		} else {
+			config.Logger.Warn(err)
+		}
 		return
 	}
 
--- a/sensorprocess/sensorprocess.go
+++ b/sensorprocess/sensorprocess.go
@@ -23,7 +23,6 @@ type Config struct {
 	Timeout          time.Duration
 	Logger           golog.Logger
 	TelemetryEnabled bool
-	jobDone          bool
 }
 
 // Start polls the lidar to get the next sensor reading and adds it to the cartofacade.
@@ -35,9 +34,11 @@ func Start(
 	for {
 		select {
 		case <-ctx.Done():
-			return config.jobDone
+			return false
 		default:
-			addSensorReading(ctx, config)
+			if jobDone := addSensorReading(ctx, config); jobDone {
+				return true
+			}
 		}
 	}
 }
@@ -46,18 +47,16 @@ func Start(
 func addSensorReading(
 	ctx context.Context,
 	config Config,
-) {
+) bool {
 	tsr, err := config.Lidar.TimedSensorReading(ctx)
 	if err != nil {
 		if strings.Contains(err.Error(), "reached end of dataset") {
-			if !config.jobDone {
-				config.Logger.Warn(err)
-				config.jobDone = true
-			}
+			config.Logger.Warn(err)
+			return true
 		} else {
 			config.Logger.Warn(err)
 		}
-		return
+		return false
 	}
 
 	if tsr.Replay {
@@ -66,6 +65,7 @@ func addSensorReading(
 		timeToSleep := addSensorReadingFromLiveReadings(ctx, tsr.Reading, tsr.ReadingTime, config)
 		time.Sleep(time.Duration(timeToSleep) * time.Millisecond)
 	}
+	return false
 }
 
 // addSensorReadingFromReplaySensor adds a reading from a replay sensor to the cartofacade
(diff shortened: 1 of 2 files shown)


viam-cartographer.go
```diff
@@ -143,7 +143,10 @@ func initSensorProcess(cancelCtx context.Context, cartoSvc *CartographerService)
 	cartoSvc.sensorProcessWorkers.Add(1)
 	go func() {
 		defer cartoSvc.sensorProcessWorkers.Done()
-		sensorprocess.Start(cancelCtx, spConfig)
+		if jobDone := sensorprocess.Start(cancelCtx, spConfig); jobDone {
+			cartoSvc.jobDone = true
+			cartoSvc.cancelSensorProcessFunc()
```
kim-mishra, line 148: Is there a harm in calling this function here if it is also called in close? Should `cartoSvc.cancelSensorProcessFunc()` only be called in close if jobDone is false?

viam-cartographer.go
```diff
@@ -747,9 +750,24 @@ func (cartoSvc *CartographerService) getNextDataPoint(ctx context.Context, lidar
...
+}
+
+func (cartoSvc *CartographerService) doCommandModularizationV2(
+	ctx context.Context,
+	req map[string]interface{},
+) (map[string]interface{}, error) {
+	if _, ok := req["job_done"]; ok {
+		return map[string]interface{}{"job_done": cartoSvc.jobDone}, nil
+	}
+
```
kim-mishra, line 770: I thought it would be useful to keep full mod changes within the feature flag for organizational purposes but I can take it outside if that seems better

sensorprocess/sensorprocess.go
```diff
@@ -44,11 +47,11 @@ func Start(
 func addSensorReading(
 	ctx context.Context,
 	config Config,
-) {
+) bool {
 	tsr, err := config.Lidar.TimedSensorReading(ctx)
 	if err != nil {
 		config.Logger.Warn(err)
-		return
+		return strings.Contains(err.Error(), "reached end of dataset")
```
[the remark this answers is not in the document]
nicksanford: Lets confirm that this error is indeed what replay sensor will return when we have reached the end of a fineite dataset i.e. one with a start & end time, and that we will get a different error if the replay sensor has an infinite dataset (i.e. no end time).

viam-cartographer.go
```diff
@@ -143,7 +143,10 @@ func initSensorProcess(cancelCtx context.Context, cartoSvc *CartographerService)
 	cartoSvc.sensorProcessWorkers.Add(1)
 	go func() {
 		defer cartoSvc.sensorProcessWorkers.Done()
-		sensorprocess.Start(cancelCtx, spConfig)
+		if jobDone := sensorprocess.Start(cancelCtx, spConfig); jobDone {
+			cartoSvc.jobDone = true
```
nicksanford, line 147: Lets make `jobDone` an atomic bool, so that we don't need to worry bout data races between setting this & reading it from DoCommand.

nicksanford: https://pkg.go.dev/sync/atomic#Bool

viam-cartographer.go
```diff
@@ -750,6 +753,21 @@ func (cartoSvc *CartographerService) DoCommand(ctx context.Context, req map[stri
 		cartoSvc.logger.Warn("DoCommand called after closed")
 		return nil, ErrClosed
 	}
+
+	if cartoSvc.modularizationV2Enabled {
```
nicksanford, line 757: lets remove this if statement please

viam-cartographer.go
nicksanford: I'd recommend testing reconfiguring after jobdone is set to true to confirm this doesn't affect the Close() function, but assuming it doesn't this looks great.

sensorprocess/sensorprocess_test.go
```diff
@@ -243,6 +243,41 @@ func TestAddSensorReadingLiveReadings(t *testing.T) {
 	})
 }
 
+func invalidSensorTestHelper(
```
nicksanford, line 246: Can we please also add the following tests for Start:
1. When the context is cancelled before the error is hit
2. When addSensorData returns false

viam-cartographer.go
```diff
@@ -241,8 +245,10 @@ func New(
 		cartoFacadeTimeout:            cartoFacadeTimeout,
 		localizationMode:              mapRateSec == 0,
 		mapTimestamp:                  time.Now().UTC(),
-		jobDone:                       false,
 	}
+
+	cartoSvc.jobDone.Store(false)
```
nicksanford, line 250: I think you can remove this given that the zero value of an atomic bool is already false, so this should already be set to false when the `CartographerService{}` struct is created above.

viam-cartographer.go
```diff
@@ -750,6 +756,11 @@ func (cartoSvc *CartographerService) DoCommand(ctx context.Context, req map[stri
 		cartoSvc.logger.Warn("DoCommand called after closed")
 		return nil, ErrClosed
 	}
+
+	if _, ok := req["job_done"]; ok {
+		return map[string]interface{}{"job_done": cartoSvc.jobDone.Load()}, nil
```
nicksanford, line 761: Have you manually tested through the `viam` cli that this works as expected both before & after the job is done?

nicksanford: sorry, just saw this in the PR description. Answer was yes

kim-mishra: Yes, heres the command:

```
~ viam --base-url=https://pr-2017-appmain-bplesliplq-uc.a.run.app:443 robot part run --robot kims-robot --location "First Location" --organization "Nils's Org" -part kims-robot-main  -d '{"name": "slam-svc", "command": {"job_done": "true"}}' viam.service.slam.v1.SLAMService.DoCommand
```

Here's the output before the log:
```
{
  "result": {
      "job_done": false
    }
}
```

after the log:
```
{
  "result": {
      "job_done": true
    }
}
```

and after reconfiguring fails:
```
ERROR:
  Code: Unknown
  Message: resource (viam:slam:cartographer) is closed
```

sensorprocess/sensorprocess_test.go
kim-mishra: added!

nicksanford (approved):

PR #208, opened 2023-07-20
RSDK-4074 - match end of dataset error

kim-mishra:
use errors.Is for error in replaypcd package we want to match against instead of string.contains and maintaining the sync between the strings


go.mod
```diff
@@ -11,10 +11,10 @@ require (
 	github.com/rhysd/actionlint v1.6.24
 	github.com/viamrobotics/gostream v0.0.0-20230609200515-c5d67c29ed25
 	go.opencensus.io v0.24.0
-	go.viam.com/api v0.1.145
-	go.viam.com/rdk v0.4.0-rc0
+	go.viam.com/api v0.1.156
+	go.viam.com/rdk v0.5.1-0.20230719205427-c10eab2aa624
```
kim-mishra, line 15: From what I can gather, latest seems to be the released version, as the docs are tagged with latest [here](https://pkg.go.dev/go.viam.com/rdk), but this does not contain my changes made to rdk yesterday so instead I ran `go get go.viam.com/rdk@main`. Is this the correct version we want to point to?

nicksanford (approved):

PR #215, opened 2023-07-21
RSDK-4064 - update mock lidar artifact dataset

kim-mishra:
update mock lidar artifact dataset to reflect data format that we use in the code

### DIFF
--- a/.artifact/tree.json
+++ b/.artifact/tree.json
@@ -2,64 +2,64 @@
   "viam-cartographer": {
     "mock_lidar": {
       "0.pcd": {
-        "hash": "ff60d0b3ae61004a1ba2031eb7c6b832",
-        "size": 39791
+        "hash": "fc063341f9586fcddc69df23f2870de7",
+        "size": 14515
       },
       "1.pcd": {
-        "hash": "db5bd806efd2bbb61c161d7c609395f9",
-        "size": 39772
+        "hash": "6224fee6b391f47065bfa47888d45653",
+        "size": 14608
       },
       "10.pcd": {
-        "hash": "ae63f9547ae2811edbad8fe09249dc41",
-        "size": 39848
+        "hash": "7e88a8c8f2984a2f80e7520750ef8ed6",
+        "size": 11819
       },
       "11.pcd": {
-        "hash": "b63f819b53bf30461513e20fec59d370",
-        "size": 40302
+        "hash": "920a7d31cc9e0d0ed653e324cd7b9c0b",
+        "size": 11504
       },
       "12.pcd": {
-        "hash": "d5b1504da753c256648bf12cc290c449",
-        "size": 39988
+        "hash": "b93d4aeca9afeeaa8d957a064046d038",
+        "size": 13046
       },
       "13.pcd": {
-        "hash": "3a928b03dbe1f805e2865ae1c8824058",
-        "size": 40042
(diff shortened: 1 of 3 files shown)


integration_test.go
```diff
@@ -52,10 +52,12 @@ func testCartographerMap(t *testing.T, svc slam.Service, localizationMode bool)
...
 func testCartographerPosition(t *testing.T, svc slam.Service, expectedComponentRef string) {
-	expectedPosOSX := r3.Vector{X: -4.534156132500952, Y: 9.800089705182817, Z: 0}
-	expectedPosLinux := r3.Vector{X: -3.0874594959315647, Y: 4.502970388777862, Z: 0}
+	expectedPosOSX := r3.Vector{X: 155.7488316264227, Y: -90.25868252233964, Z: 0}
+	expectedPosLinux := r3.Vector{X: 158.79903385710674, Y: -77.01514065531592, Z: 0}
 	tolerancePos := 0.001
-	expectedOri := &spatialmath.R4AA{Theta: 0.0018812019022632894, RX: 0, RY: 0, RZ: -1}
+
+	expectedOriOSX := &spatialmath.R4AA{Theta: 1.5465081272043815, RX: 0, RY: 0, RZ: 1}
+	expectedOriLinux := &spatialmath.R4AA{Theta: 0.3331667853231311, RX: 0, RY: 0, RZ: 1}
```
kim-mishra, line 60: After updating the dataset the orientations also differed slightly between architectures. The theta value's units are degrees as per the [proto](https://github.com/viamrobotics/api/blob/69f8be1dc959587244c71e044d54b004be2d9c77/common/v1/common.pb.go#L324C10-L324C10). This seems to be on a similar order of magnitude of error as a few millimeters like in the position, but want to make sure that orientation varying slightly does not mean we need to investigate any further

.artifact/tree.json
```diff
@@ -1,5 +1,5 @@
 {
-  "viam-cartographer": {
+  "slam": {
```
kim-mishra, line 2: I am not exactly sure why this happened, it auto generated after running artifact push and only making changes to the viam-cartographer/mock_lidar directory. Did I miss some clean up that needed to be done?

nicksanford: I think that this would only happen if you had lidar readings at `.artifact/data/slam/mock_lidar/`

They should be deleted

integration_test.go
nicksanford: I think its unsurprising that the orientation differs slightly on mac & linux in the because the position differs between platforms.

This can stay as is.

testhelper/testhelper.go
```diff
@@ -145,7 +145,7 @@ func IntegrationLidarTimedSensor(
 
 	ts.TimedSensorReadingFunc = func(ctx context.Context) (s.TimedSensorReadingResponse, error) {
 		readingTime = readingTime.Add(sensorReadingInterval)
-		t.Logf("TimedSenorReading Mock i: %d, closed: %v, readingTime: %s\n", i, closed, readingTime.String())
+		t.Logf("TimedSensorReading Mock i: %d, closed: %v, readingTime: %s\n", i, closed, readingTime.String())
```
nicksanford, line 148: ty for the typo fix

nicksanford (approved):

PR #235, opened 2023-08-02
[RSDK-4137] Config changes for IMU Support

PStrutz:
- Add IMU config parameters in Go code.
- Add IMU into cartographer struct and cartofacade config
- Add IMU as an optional parameter in GetOptionalParameters
- Add tests for IMU config

### DIFF
--- a/cartofacade/capi.go
+++ b/cartofacade/capi.go
@@ -87,6 +87,7 @@ const (
 // CartoConfig contains config values from app
 type CartoConfig struct {
 	Camera             string
+	MovementSensor     string
 	MapRateSecond      int
 	DataDir            string
 	ComponentReference string
@@ -323,6 +324,7 @@ func toLidarConfig(lidarConfig LidarConfig) (C.viam_carto_LIDAR_CONFIG, error) {
 func getConfig(cfg CartoConfig) (C.viam_carto_config, error) {
 	vcc := C.viam_carto_config{}
 	vcc.camera = goStringToBstring(cfg.Camera)
+	vcc.movement_sensor = goStringToBstring(cfg.MovementSensor)
 
 	lidarCfg, err := toLidarConfig(cfg.LidarConfig)
 	if err != nil {
--- a/config/config.go
+++ b/config/config.go
@@ -17,6 +17,7 @@ func newError(configError string) error {
 // Config describes how to configure the SLAM service.
 type Config struct {
 	Camera                map[string]string `json:"camera"`
+	MovementSensor        map[string]string `json:"movement_sensor"`
 	ConfigParams          map[string]string `json:"config_params"`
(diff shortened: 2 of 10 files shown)


cartofacade/capi.go
```diff
@@ -323,6 +324,7 @@ func toLidarConfig(lidarConfig LidarConfig) (C.viam_carto_LIDAR_CONFIG, error) {
 func getConfig(cfg CartoConfig) (C.viam_carto_config, error) {
 	vcc := C.viam_carto_config{}
 	vcc.camera = goStringToBstring(cfg.Camera)
+	vcc.movement_sensor = goStringToBstring(cfg.MovementSensor)
```
EshaMaharishi, line 327: did you figure out what was missing here?

config/config.go
```diff
@@ -91,10 +94,24 @@ func GetOptionalParameters(config *Config, defaultLidarDataRateMsec, defaultMapR
 		} else {
 			lidarDataFreqHz, err := strconv.Atoi(strCameraDataFreqHz)
 			if err != nil {
-				return 0, 0, newError("camera[data_frequency_hz] must only contain digits")
+				return 0, "", 0, 0, newError("camera[data_frequency_hz] must only contain digits")
 			}
 			lidarDataRateMsec = 1000 / lidarDataFreqHz
 		}
+		imuName, ok = config.MovementSensor["name"]
```
EshaMaharishi, line 101: instead of 'ok', let's say 'exists', since as a reader it's surprising to see an `if ok` without an `else` to handle the "not ok" case

config/config.go
```diff
@@ -91,10 +94,24 @@ func GetOptionalParameters(config *Config, defaultLidarDataRateMsec, defaultMapR
...
+				logger.Debugf("config did not provide movement_sensor[data_frequency_hz], setting to default value of %d", 1000/defaultIMUDataRateMsec)
+			} else {
+				imuDataFreqHz, err := strconv.Atoi(strMovementSensorDataFreqHz)
+				if err != nil {
+					return 0, "", 0, 0, newError("movement_sensor[data_frequency_hz] must only contain digits")
+				}
+				imuDataRateMsec = 1000 / imuDataFreqHz
+			}
+		}
+		// should IMU data rate be set to 0 if no IMU is configured?
```
EshaMaharishi, line 114: Ah, I see what you mean now about whether to leave it as the default rate or set it to 0. I think it can be left as the default rate.

EshaMaharishi (approved):
LGTM mod question about if you were able to get the code to auto-generate

cartofacade/capi.go
PStrutz: Yep! Had to compile everything by running make build

EshaMaharishi:
Ah you have a memory leak. We can talk about that and should fix it before merging

PR #236, opened 2023-08-03
RSDK-4088 - use data_rate_msec to determine if sensor is in online or offline mode

kim-mishra:
- passes param isLive to sensor process and uses that to choose which `addSensorReading` func to call instead of basing this on sensor type
- updates tests to reflect changes

### DIFF
--- a/sensorprocess/sensorprocess.go
+++ b/sensorprocess/sensorprocess.go
@@ -53,7 +53,9 @@ func addSensorReading(
 	tsr, err := config.Lidar.TimedSensorReading(ctx)
 	if err != nil {
 		config.Logger.Warn(err)
-		return strings.Contains(err.Error(), replaypcd.ErrEndOfDataset.Error())
+		isEndOfDataset := strings.Contains(err.Error(), replaypcd.ErrEndOfDataset.Error())
+		jobDone := !isOnline && isEndOfDataset
+		return jobDone
 	}
 	if isOnline {
 		timeToSleep := addSensorReadingOnline(ctx, tsr.Reading, tsr.ReadingTime, config)
--- a/sensorprocess/sensorprocess.go
+++ b/sensorprocess/sensorprocess.go
@@ -30,14 +30,14 @@ type Config struct {
 func Start(
 	ctx context.Context,
 	config Config,
-	isLive bool,
+	isOnline bool,
 ) bool {
 	for {
 		select {
 		case <-ctx.Done():
 			return false
 		default:
-			if jobDone := addSensorReading(ctx, config, isLive); jobDone {
+			if jobDone := addSensorReading(ctx, config, isOnline); jobDone {
 				return true
 			}
 		}
@@ -48,14 +48,14 @@ func Start(
 func addSensorReading(
 	ctx context.Context,
 	config Config,
-	isLive bool,
+	isOnline bool,
 ) bool {
 	tsr, err := config.Lidar.TimedSensorReading(ctx)
 	if err != nil {
 		config.Logger.Warn(err)
 		return strings.Contains(err.Error(), replaypcd.ErrEndOfDataset.Error())
 	}
-	if isLive {
+	if isOnline {
 		timeToSleep := addSensorReadingOnline(ctx, tsr.Reading, tsr.ReadingTime, config)
 		time.Sleep(time.Duration(timeToSleep) * time.Millisecond)
 	} else {
--- a/sensorprocess/sensorprocess.go
+++ b/sensorprocess/sensorprocess.go
@@ -37,7 +37,7 @@ func Start(
 		case <-ctx.Done():
 			return false
 		default:
-			if jobDone := addSensorReading(ctx, config, isOnline); jobDone {
+			if jobDone := addSensorReading(ctx, config); jobDone {
 				return true
 			}
 		}
@@ -48,20 +48,23 @@ func Start(
 func addSensorReading(
 	ctx context.Context,
 	config Config,
-	isOnline bool,
 ) bool {
 	tsr, err := config.Lidar.TimedSensorReading(ctx)
 	if err != nil {
 		config.Logger.Warn(err)
-		isEndOfDataset := strings.Contains(err.Error(), replaypcd.ErrEndOfDataset.Error())
-		jobDone := isEndOfDataset && !isOnline
-		return jobDone
+		// config.LidarDataRateMsec == 0 means we are in offline mode
+		if config.LidarDataRateMsec == 0 {
+			return strings.Contains(err.Error(), replaypcd.ErrEndOfDataset.Error())
+		} else {
+			return false
+		}
 	}
-	if isOnline {
+	// config.LidarDataRateMsec == 0 means we are in offline mode
+	if config.LidarDataRateMsec == 0 {
+		addSensorReadingOffline(ctx, tsr.Reading, tsr.ReadingTime, config)
+	} else {
 		timeToSleep := addSensorReadingOnline(ctx, tsr.Reading, tsr.ReadingTime, config)
 		time.Sleep(time.Duration(timeToSleep) * time.Millisecond)
-	} else {
-		addSensorReadingOffline(ctx, tsr.Reading, tsr.ReadingTime, config)
 	}
 	return false
 }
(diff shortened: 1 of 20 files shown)


sensorprocess/sensorprocess.go
```diff
@@ -30,13 +30,14 @@ type Config struct {
 func Start(
 	ctx context.Context,
 	config Config,
+	isOnline bool,
```
kim-mishra, line 33: Should we call this `isOnline` or `isLive`? 

To me, "online" is the opposite of "offline" and "live" is the opposite of a  "replay sensor" but happy to change if live makes more sense

sensorprocess/sensorprocess_test.go
```diff
@@ -358,60 +422,12 @@ func TestAddSensorReading(t *testing.T) {
...
-			test.That(t, call.sensorName, test.ShouldResemble, "good_lidar")
-			// the lidar test fixture happens to always return the same pcd currently
-			// in reality it could be a new pcd every time
-			test.That(t, call.currentReading, test.ShouldResemble, expectedPCD)
-			test.That(t, call.timeout, test.ShouldEqual, config.Timeout)
-		}
-		test.That(t, calls[0].readingTimestamp.Before(calls[1].readingTimestamp), test.ShouldBeTrue)
-		test.That(t, calls[1].readingTimestamp.Before(calls[2].readingTimestamp), test.ShouldBeTrue)
+	t.Run("online lidar adds sensor reading once and ignores errors", func(t *testing.T) {
+		onlineModeTestHelper(ctx, t, config, cf, "replay_lidar")
```
kim-mishra, line 430: helper func described above

sensorprocess/sensorprocess_test.go
```diff
@@ -358,60 +422,12 @@ func TestAddSensorReading(t *testing.T) {
...
-
-		jobDone = addSensorReading(ctx, config)
-		test.That(t, len(calls), test.ShouldEqual, 2)
-		test.That(t, jobDone, test.ShouldBeFalse)
-
-		jobDone = addSensorReading(ctx, config)
-		test.That(t, len(calls), test.ShouldEqual, 3)
-		test.That(t, jobDone, test.ShouldBeFalse)
+	t.Run("online replay lidar adds sensor reading once and ignores errors", func(t *testing.T) {
+		onlineModeTestHelper(ctx, t, config, cf, "good_lidar")
```
kim-mishra, line 426: helper func described above

sensorprocess/sensorprocess_test.go
```diff
@@ -236,18 +236,91 @@ func TestAddSensorReadingLiveReadings(t *testing.T) {
...
+	if cam == "good_lidar" {
+		test.That(t, calls[0].readingTimestamp.Before(calls[1].readingTimestamp), test.ShouldBeTrue)
+		test.That(t, calls[1].readingTimestamp.Before(calls[2].readingTimestamp), test.ShouldBeTrue)
+	} else if cam == "replay_lidar" {
+		readingTime, err := time.Parse(time.RFC3339Nano, s.TestTime)
+		test.That(t, err, test.ShouldBeNil)
+		test.That(t, calls[0].readingTimestamp.Equal(readingTime), test.ShouldBeTrue)
+	} else {
+		t.Errorf("no timestamp tests provided for %v", cam)
+	}
```
kim-mishra, line 314: This is the only part that was changed so that the test could be generalized across replay & live lidars

sensorprocess/sensorprocess.go
```diff
@@ -47,24 +48,27 @@ func Start(
...
 ) bool {
 	tsr, err := config.Lidar.TimedSensorReading(ctx)
 	if err != nil {
 		config.Logger.Warn(err)
-		return strings.Contains(err.Error(), replaypcd.ErrEndOfDataset.Error())
+		isEndOfDataset := strings.Contains(err.Error(), replaypcd.ErrEndOfDataset.Error())
+		jobDone := isEndOfDataset && !isOnline
+		return jobDone
```
kim-mishra, line 58: This change makes it so we do not end the sensor process if we are in offline mode - should we sleep here?

sensorprocess/sensorprocess.go
```diff
@@ -47,24 +48,25 @@ func Start(
...
+		addSensorReadingOffline(ctx, tsr.Reading, tsr.ReadingTime, config)
 	}
 	return false
 }
 
-// addSensorReadingFromReplaySensor adds a reading from a replay sensor to the cartofacade
+// addSensorReadingOffline adds a reading to the cartofacade
 // retries on error.
-func addSensorReadingFromReplaySensor(ctx context.Context, reading []byte, readingTime time.Time, config Config) {
+func addSensorReadingOffline(ctx context.Context, reading []byte, readingTime time.Time, config Config) {
```
EshaMaharishi, line 71: What do you think of calling this tryAddSensorReadingUntilSuccess?

Also, it might be possible to write this in terms of tryAddSensorReading (but not a big deal if not or if it ends up looking less clear that way)

sensorprocess/sensorprocess.go
```diff
@@ -85,9 +87,9 @@ func addSensorReadingFromReplaySensor(ctx context.Context, reading []byte, readi
 	}
 }
 
-// addSensorReadingFromLiveReadings adds a reading from a live lidar to the carto facade
+// addSensorReadingOnline adds a reading to the carto facade
 // does not retry.
-func addSensorReadingFromLiveReadings(ctx context.Context, reading []byte, readingTime time.Time, config Config) int {
+func addSensorReadingOnline(ctx context.Context, reading []byte, readingTime time.Time, config Config) int {
```
EshaMaharishi, line 94: What do you think of calling this tryAddSensorReading?

sensorprocess/sensorprocess.go
EshaMaharishi: (see other comment)

viam_cartographer.go
```diff
@@ -127,9 +127,11 @@ func initSensorProcess(cancelCtx context.Context, cartoSvc *CartographerService)
 	}
 
 	cartoSvc.sensorProcessWorkers.Add(1)
+	isOnline := cartoSvc.lidarDataRateMsec != 0
```
EshaMaharishi, line 130: Since the data rate msec is already being passed via spConfig, can we just check the data rate msec directly in Start?

Then we don't have to worry about the name (isOnline vs. live), and can just add a comment that when the data rate msec is 0, we assume the user wants to be in "offline" mode and ensure every scan gets processed by cartographer

EshaMaharishi (approved):
LGTM with suggestions

Neatly done!

sensorprocess/sensorprocess.go
EshaMaharishi: Oops, I missed this - Can we check data rate msec from `config` here and end the sensor process if we get back ErrEndOfDataset and the data rate msec is 0?

kim-mishra: `isOnline` comes from data_rate_msec here. Do you want the whole config to be passed in instead?

viam_cartographer.go
kim-mishra: ah got it! changing

sensorprocess/sensorprocess.go
kim-mishra: nvm got it!

sensorprocess/sensorprocess.go
kim-mishra: updating!

kim-mishra: I think writing ` tryAddSensorReadingUntilSuccess` in terms of ` tryAddSensorReading` ends up being a little less clear because really they are both wrappers around `CartoFacade.AddSensorReading`

EshaMaharishi: ok, sounds good!

sensorprocess/sensorprocess.go
```diff
@@ -30,6 +30,7 @@ type Config struct {
 func Start(
 	ctx context.Context,
 	config Config,
+	isOnline bool,
```
EshaMaharishi, line 33: can we remove this parameter now?

viam_cartographer.go
EshaMaharishi, line 130: can we remove the changes to this file?

kim-mishra: fixed thanks for catching! i am surprised i didn't get an error for isOnline not being used in Start

EshaMaharishi (approved):

PR #237, opened 2023-08-04
RSDK-4345 - add feature flag for cloud story

kim-mishra:
# Changes
* adds feature flag to put cloud story changes behind. Feature flag will be provided through the config as `cloud_story_enabled`
* adds config changes for `existing_map`, `enable_mapping`, and `run_slam`
* validates that `existing_map` has been provided
* adds parsing & tests for new config params
* does not create a new getOptionalParameters func

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -39,90 +39,50 @@ var (
 // Validate creates the list of implicit dependencies.
 func (config *Config) Validate(path string) ([]string, error) {
 	cameraName := ""
+
+	if config.IMUIntegrationEnabled {
+		var ok bool
+		cameraName, ok = config.Camera["name"]
+		if !ok {
+			return nil, utils.NewConfigValidationError(path, errCameraMustHaveName)
+		}
+		dataFreqHz, ok := config.Camera["data_frequency_hz"]
+		if ok {
+			dataFreqHz, err := strconv.Atoi(dataFreqHz)
+			if err != nil {
+				return nil, errors.New("camera[data_frequency_hz] must only contain digits")
+			}
+			if dataFreqHz < 0 {
+				return nil, errors.New("cannot specify camera[data_frequency_hz] less than zero")
+			}
+		}
+	} else {
+		if config.Sensors == nil || len(config.Sensors) < 1 {
+			return nil, utils.NewConfigValidationError(path, errSensorsMustNotBeEmpty)
+		}
+		cameraName = config.Sensors[0]
+
+		if config.DataRateMsec < 0 {
+			return nil, errors.New("cannot specify data_rate_msec less than zero")
+		}
+	}
+
+	if config.ConfigParams["mode"] == "" {
+		return nil, utils.NewConfigValidationFieldRequiredError(path, "config_params[mode]")
+	}
+
 	if config.CloudStoryEnabled {
-		return config.ValidateCloudStoryEnabled(path)
-	}
-
-	if config.IMUIntegrationEnabled {
-		var ok bool
-		cameraName, ok = config.Camera["name"]
-		if !ok {
-			return nil, utils.NewConfigValidationError(path, errCameraMustHaveName)
-		}
-		dataFreqHz, ok := config.Camera["data_frequency_hz"]
-		if ok {
-			dataFreqHz, err := strconv.Atoi(dataFreqHz)
-			if err != nil {
-				return nil, errors.New("camera[data_frequency_hz] must only contain digits")
-			}
-			if dataFreqHz < 0 {
-				return nil, errors.New("cannot specify camera[data_frequency_hz] less than zero")
-			}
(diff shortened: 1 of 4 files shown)


config/config.go
```diff
@@ -78,6 +87,49 @@ func (config *Config) Validate(path string) ([]string, error) {
...
+		}
+	}
+
+	if config.ConfigParams["mode"] == "" {
+		return nil, utils.NewConfigValidationFieldRequiredError(path, "config_params[mode]")
+	}
+
+	if config.ExistingMap == "" {
+		return nil, utils.NewConfigValidationFieldRequiredError(path, "existing_map")
+	}
```
kim-mishra, line 126: Will `existing_map` this be required?

config/config.go
```diff
@@ -78,6 +87,49 @@ func (config *Config) Validate(path string) ([]string, error) {
...
+
+	deps := []string{cameraName}
+
+	return deps, nil
+}
+
 // GetOptionalParameters sets any unset optional config parameters to the values passed to this function,
 // and returns them.
 func GetOptionalParameters(config *Config, defaultLidarDataRateMsec, defaultIMUDataRateMsec, defaultMapRateSec int, logger golog.Logger,
```
kim-mishra, line 135: This will need to be changed as well, but i am not sure if it is worth changing until more of the cartographer changed are made for cloud story. Thoughts?

config/config_test.go
```diff
@@ -10,6 +10,39 @@ import (
 	"go.viam.com/utils"
 )
 
+const CloudStoryEnabled = "cloud_story_enabled"
+
+func configWithoutRequiredFieldsTestHelper(
```
kim-mishra, line 15: This was pulled out from an existing test to use to test with & without the feature flag

config/config.go
JohnN193: for now yes, `ExistingMap` should effectively be the same as `DataDirectory`. in https://viam.atlassian.net/browse/RSDK-4115 it will change to only be required for `EnableMapping=false`

config/config.go
JohnN193: fine with leaving this as is until the relevant tickets are implemented

config/config.go
```diff
@@ -34,6 +39,10 @@ var (
 // Validate creates the list of implicit dependencies.
 func (config *Config) Validate(path string) ([]string, error) {
 	cameraName := ""
+	if config.CloudStoryEnabled {
+		return config.ValidateCloudStoryEnabled(path)
```
JohnN193, line 43: does this need to return here? can we just check our params exist then run the remainder of validate?

kim-mishra: updated to reflect changes we talked about offline

config/config_test.go
```diff
@@ -10,25 +11,25 @@ import (
...
-		_, err := newConfig(cfgService)
-		test.That(t, err, test.ShouldBeError, newError("error validating \"services.slam.attributes.fake\": \"sensors\" must not be empty"))
-	})
-
+func testValidateTesthelper(
+	t *testing.T,
+	logger golog.Logger,
+	testCfgPath string,
+	cloudStoryEnabled bool,
+) {
```
JohnN193, line 19: is there any way to get the `t.Run`s to include if the feature flag was enabled? 

My naive  approach being 
```
withFeatureFlag = ""
if cloudStoryEnabled{
   withFeatureFlag = "with cloudStoryEnabled"
}
t.Run("Simplest valid config" + withFeatureFlag, func(t *testing.T)
```

config/config_test.go
```diff
@@ -87,23 +106,43 @@ func TestValidate(t *testing.T) {
...
+
+	t.Run("Empty config", func(t *testing.T) {
+		model := resource.DefaultModelFamily.WithModel("test")
+		cfgService := resource.Config{Name: "test", API: slam.API, Model: model}
+		_, err := newConfig(cfgService)
+		test.That(t, err, test.ShouldBeError, newError("error validating \"services.slam.attributes.fake\": \"sensors\" must not be empty"))
+	})
+
+	testValidateTesthelper(t, logger, testCfgPath, false)
+	testValidateTesthelper(t, logger, testCfgPath, true)
```
JohnN193, line 135: [q] so `testValidateHelper` goes through all tests, both with and without the imu feature flag?

JohnN193 (approved):
LGTM, would just want a way to see which tests have the feature flag enabled within `testValidateTesthelper`

config/config_test.go
kim-mishra: oh yes definitely will add this in!

config/config_test.go
kim-mishra: `testValidateHelper` goes through all of the tests although it looks like there were no existing tests tests where imuIntegrationEnabled was set to true

PR #239, opened 2023-08-07
fix to not overwrite data_rate_msec when providing 0

kim-mishra:
Change `dataRateMsec` to be *int so we can check if it is null instead of 0 this way we can use `data_rate_msec = 0` to mean offline mode

### DIFF
--- a/viam_cartographer_test.go
+++ b/viam_cartographer_test.go
@@ -24,10 +24,10 @@ import (
 )
 
 const (
-	testExecutableName = "true" // the program "true", not the boolean value
-	testDataFreqHz     = "5"
-  testIMUDataFreqHz   = "20"
-	testDataRateMsec    = 200
+	testExecutableName  = "true" // the program "true", not the boolean value
+	testDataFreqHz      = "5"
+	testIMUDataFreqHz   = "20"
+	testLidarDataFreqHz = "5"
 )
 
 var (
@@ -35,7 +35,7 @@ var (
 	_zeroInt         = 0
 	_zeroTime        = time.Time{}
 	testDataRateMsec = 200
-  _true          = true
+	_true            = true
 )
 
 func TestNew(t *testing.T) {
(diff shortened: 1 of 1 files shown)


sensorprocess/sensorprocess.go
```diff
@@ -66,6 +66,7 @@ func addSensorReading(
 	} else {
 		timeToSleep := tryAddSensorReading(ctx, tsr.Reading, tsr.ReadingTime, config)
 		time.Sleep(time.Duration(timeToSleep) * time.Millisecond)
+		config.Logger.Infof("sleep for %s", time.Duration(timeToSleep))
```
JohnN193, line 69: can this be a debug log

config/config.go
```diff
@@ -62,7 +62,7 @@ func (config *Config) Validate(path string) ([]string, error) {
 		}
 		cameraName = config.Sensors[0]
 
-		if config.DataRateMsec < 0 {
+		if config.DataRateMsec != nil && *config.DataRateMsec < 0 {
```
JohnN193, line 65: will the IMU run into a similar issue or is it fine to leave as is?

JohnN193 (approved):
LGTM just a nit and a question

config/config.go
kim-mishra: This is a problem for lidars becauses  `data_rate_msec m = 0` needs to be differentiated from  `data_rate_msec m = nill` so that we can determine if we are in offline or online. I am not sure if there will be an alternate sensor process for imus or if we can rely on the lidarDataRateMsec to close both. @PStrutz Do you know?

PStrutz: There will be a separate sensor process, but I can configure it to check the lidar data rate to determine offline/online mode

kim-mishra: I think that makes sense because we do not want to define the mode in multiple places @JohnN193 what do you think?

JohnN193: yeah that makes sense, but for this pr replay cameras will still be broken for the imu integration feature flag. If we are okay with that & letting @PStrutz handling it when she gets to replay for imus then thats fine with me

kim-mishra: @JohnN193 after talking with @EshaMaharishi it sounds like `data_frequency_hz` when IMU ff is on is the new way to represent `data_rate_msec`. So I updated the checks to make it so if that is not provided we set it to the default value, if it is provided and not zero same behavior as before and if it is provided and is 0 we just set the `dataRateMsec` to 0. Let me know what you think now

cc @PStrutz

PR #240, opened 2023-08-07
RSDK-4362 - remove map rate sec and stop saving maps

kim-mishra:
## High Level Cloud Story Config Changes overview
1.  add `enable_mapping` to config. If `cloud_story_enabled` == true: use `enable_mapping=false` in place of `map_rate_sec = 0` to determine if we are in `LOCALIZING` mode or not
2. **if cloud story is enabled, do not validate `map_rate_sec`. Remove logic to save internal state to the `data_dir`**
3. If cloud story is enabled stop validating `data_dir`, add `existing_map` to optional parameters. Use `existing_state` to get any existing map 

**This PR implements number 2**

* Do not validate `map_rate_sec` if cloud story enabled
* Do not call cartofacade `start()` or `stop()` if cloud story is enabled (these handle saving internal state)

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -151,10 +151,10 @@ func GetOptionalParameters(config *Config, defaultLidarDataRateMsec, defaultIMUD
 			optionalConfigParams.EnableMapping = config.CloudStoryEnabled
 		}
 		return optionalConfigParams, nil
-	} else {
-		if config.EnableMapping != nil && *config.EnableMapping {
-			logger.Warn("enable_mapping set to true while cloud_story_enabled = false will not change any behavior")
-		}
+	}
+
+	if config.EnableMapping != nil && *config.EnableMapping {
+		logger.Warn("enable_mapping set to true while cloud_story_enabled = false will not change any behavior")
 	}
 
 	if config.MapRateSec == nil {
--- a/config/config.go
+++ b/config/config.go
@@ -68,20 +68,6 @@ func (config *Config) Validate(path string) ([]string, error) {
 			}
 		}
 
-<<<<<<< HEAD
-=======
-		mapRateSec, ok := config.Camera["map_rate_sec"]
-		if ok {
-			mapRateSec, err := strconv.Atoi(mapRateSec)
-			if err != nil {
-				return nil, errors.New("camera[map_rate_sec] must only contain digits")
-			}
-			if mapRateSec < 0 {
-				return nil, errors.New("cannot specify map_rate_sec less than zero")
-			}
-		}
-
->>>>>>> ef56c5e (pulled in IMU integration tests)
 		imuName, imuExists = config.MovementSensor["name"]
 	} else {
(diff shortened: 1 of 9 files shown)


viam_cartographer.go
```diff
@@ -385,15 +390,18 @@ func initCartoFacade(ctx context.Context, cartoSvc *CartographerService) error {
...
+	if !cartoSvc.cloudStoryEnabled {
+		err = cf.Start(ctx, cartoSvc.cartoFacadeTimeout)
+		if err != nil {
+			cartoSvc.logger.Errorw("cartofacade start failed", "error", err)
+			termErr := cf.Terminate(ctx, cartoSvc.cartoFacadeTimeout)
+			if termErr != nil {
+				cartoSvc.logger.Errorw("cartofacade terminate failed", "error", termErr)
+				return termErr
+			}
+			return err
```
kim-mishra, line 403: this is where the background process to call save internal state gets called

config/config.go
```diff
@@ -161,6 +154,14 @@ func GetOptionalParameters(config *Config, defaultLidarDataRateMsec, defaultIMUD
 		if config.EnableMapping != nil && *config.EnableMapping {
 			logger.Warn("enable_mapping set to true while cloud_story_enabled = false will not change any behavior")
 		}
+		return optionalConfigParams, nil
```
kim-mishra, line 157: return params without setting `map_rate_sec` if `cloud_story_enabled`

viam_cartographer.go
```diff
@@ -454,9 +457,13 @@ func terminateCartoFacade(ctx context.Context, cartoSvc *CartographerService) er
 		cartoSvc.logger.Debug("terminateCartoFacade called when cartoSvc.cartofacade is nil")
 		return nil
 	}
-	stopErr := cartoSvc.cartofacade.Stop(ctx, cartoSvc.cartoFacadeTimeout)
-	if stopErr != nil {
-		cartoSvc.logger.Errorw("cartofacade stop failed", "error", stopErr)
+
+	var stopErr error
+	if !cartoSvc.cloudStoryEnabled {
+		stopErr := cartoSvc.cartofacade.Stop(ctx, cartoSvc.cartoFacadeTimeout)
```
JohnN193, line 463: so `cartofacade.Start` and `cartofacade.Stop` is only used for map rate sec/saving maps?

JohnN193 (approved):
LGTM just a question for my own knowledge

viam_cartographer.go
kim-mishra: yep! it calls through the CGo API into C++ function [CartoFacade::Start()](https://github.com/viamrobotics/viam-cartographer/blob/e13d374dc91f6e5cc621d7c7e5ad4028feb9541c/viam-cartographer/src/carto_facade/carto_facade.cc#L681) and in that all it does is call `StartSaveInternalState();`

PR #241, opened 2023-08-07
RSDK-3913-add-run-slam-flag

nicksanford:


config/config.go
```diff
@@ -28,7 +28,7 @@ type Config struct {
 	CloudStoryEnabled bool   `json:"cloud_story_enabled"`
 	ExistingMap       string `json:"existing_map"`
 	EnableMapping     bool   `json:"enable_mapping"`
-	RunSlam           bool   `json:"run_slam"`
+	RunSlam           *bool  `json:"run_slam"`
```
nicksanford, line 31: This needs to be a pointer so that we can distinguish between it being not in the config and being int he config but set to false. Otherwise, by default, all existing slam configs will no longer boot cartographer.

viam_cartographer.go
```diff
@@ -32,6 +32,8 @@ var (
 	cartoLib cartofacade.CartoLib
 	// ErrClosed denotes that the slam service method was called on a closed slam resource.
 	ErrClosed = errors.Errorf("resource (%s) is closed", Model.String())
+	// ErrRunSlamDisabled denotes that the slam service method was called while run_slam was set to false
+	ErrRunSlamDisabled = errors.Errorf("resource (%s) was booted with run_slam set to false", Model.String())
```
nicksanford, line 36: @JohnN193 how does this error messages look?

JohnN193: how does `"resource (%s) unavailable, configured with run_slam set to false"` sound? the initial message isn't clear to me why I am seeing the error. 

Other part I am thinking about is if we should say something about cloud slam, so users don't try to use the endpoints while cloud slam is running

nicksanford: updated. I'm not really sure how to describe in a single line error message anything about cloud slam but am happy to integrate any requests for error message changes.

JohnN193: same tbh, but fine to leave as is

config/config.go
JohnN193: shouldn't we just use the feature flag? then once the project is done we can migrate configs

JohnN193 (approved):
LGTM, only thought is whether `run_slam`'s usage should be behind the feature flag

JohnN193 (approved):

PR #245, opened 2023-08-10
RSDK-4306 Implement Properties for Mock LiDARs

jeremyrhyde:
This PR implements the Properties function for the various mock and stub lidars in viam-cartographer

JIRA Ticket: https://viam.atlassian.net/browse/RSDK-4306


sensors/sensor_mock.go
```diff
@@ -4,7 +4,7 @@ import (
 	"context"
 )
 
-// TimedLidarSensorMock represents a fake TimedLidarSensor.
+// TimedLidarSensorMock represents a fake TimedLidarSensor. (temp)
 type TimedLidarSensorMock struct {
```
jeremyrhyde, line 8: [q] Do we need to add Properties to these mock sensors? Both imu and lidar?

Note: will remove added comment before merging

sensors/test_deps.go
```diff
@@ -130,7 +130,7 @@ func getInvalidLidar() *inject.Camera {
 		return nil, transform.NewNoIntrinsicsError("")
 	}
 	cam.PropertiesFunc = func(ctx context.Context) (camera.Properties, error) {
-		return camera.Properties{}, nil
+		return camera.Properties{SupportsPCD: true}, nil
```
EshaMaharishi, line 133: probably ok to leave SupportsPCD out here, since this is for an invalid lidar

testhelper/testhelper.go
```diff
@@ -87,6 +87,11 @@ func getStubLidar(t *testing.T) *inject.Camera {
 		t.Error("TEST FAILED stub lidar Projector called")
 		return nil, transform.NewNoIntrinsicsError("")
 	}
+	cam.PropertiesFunc = func(ctx context.Context) (camera.Properties, error) {
+		return camera.Properties{
+			SupportsPCD: true,
+		}, nil
+	}
```
EshaMaharishi, line 94: do we need this here? or should we implement it with a t.Error similar to the functions above?

sensors/test_deps.go
```diff
@@ -75,7 +75,7 @@ func getWarmingUpLidar() *inject.Camera {
...
```
EshaMaharishi, line 78: Here and below, we could leave the PropertiesFunc returning an empty Properties struct unless a test actually requires it to return `SupportsPCD: true`

sensors/sensor_mock.go
EshaMaharishi: I don't think so - no need to implement mock functionality unless there's a test that actually requires it

EshaMaharishi (approved):
LGTM mod comments

sensors/sensors.go
```diff
@@ -72,18 +72,16 @@ func NewLidar(
...
 	}
 
-	// https://viam.atlassian.net/browse/RSDK-4306
-	// To be implemented once replay camera supports Properties
-	// // If there is a camera provided in the 'camera' field, we enforce that it supports PCD.
-	// properties, err := newLidar.Properties(ctx)
-	// if err != nil {
-	// 	return Lidar{}, errors.Wrapf(err, "error getting lidar camera properties %v for slam service", cameraName)
-	// }
+	// If there is a camera provided in the 'camera' field, we enforce that it supports PCD.
```
EshaMaharishi, line 75: One more thing: Could we change the name of this ticket/PR to reflect this change, which I think is the main change?

jeremyrhyde: Done: https://viam.atlassian.net/browse/RSDK-4306

EshaMaharishi: ty!

sensors/test_deps.go
PStrutz: I think since we are checking that it supports the properties upon creation of the new lidar, we do need SupportsPCD to be true here, because otherwise we would catch the unsupported properties before we get to the erroring functions

sensors/test_deps.go
PStrutz: Every time we call NewLidar on a lidar we are checking that it supports the properties, so I think we wouldn't be able to do any testing with these injected lidars if they do not fulfill the properties requirement

testhelper/testhelper.go
PStrutz: Similarly to before, we need this camera to support the properties as we are always checking them - in the functions above, we are returning an error if they are called, since we want to make sure we are using offline data. However, since the properties function is more of a sanity check, we do need this to be properly implemented

PStrutz:
Looks good to me! I responded to some of Esha's comments but I don't think there is anything actionable left to do

PStrutz (approved):
Looks good to me! I responded to some of Esha's comments but I don't think there is anything actionable left to do

PR #249, opened 2023-08-16
RSDK-2262 - use config to determine slam mode during validation

kim-mishra:
### Changes
* throw an error if config suggests `localization mode` & `offline mode`
* update tests to test above change

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -2,6 +2,7 @@
 package config
 
 import (
+	"fmt"
 	"strconv"
 	"strings"
 
@@ -43,8 +44,11 @@ type OptionalConfigParams struct {
 }
 
 var (
-	errCameraMustHaveName    = errors.New("\"camera[name]\" is required")
-	errSensorsMustNotBeEmpty = errors.New("\"sensors\" must not be empty")
+	errCameraMustHaveName              = errors.New("\"camera[name]\" is required")
+	errSensorsMustNotBeEmpty           = errors.New("\"sensors\" must not be empty")
+	errLocalizationInOfflineModeSuffix = " = 0, enable_mapping = false, and existing_map provided. localization in offline mode not supported."
+	errLocalizationInOfflineMode       = newError(fmt.Sprintf("data_rate_msec%s", errLocalizationInOfflineModeSuffix))
+	errLocalizationInOfflineModeIMU    = newError(fmt.Sprintf("camera: data_freq_hz%s", errLocalizationInOfflineModeSuffix))
 )
 
 // Validate creates the list of implicit dependencies.
@@ -161,7 +165,11 @@ func GetOptionalParameters(config *Config, defaultLidarDataRateMsec, defaultIMUD
 		if config.EnableMapping == nil {
 			logger.Debug("no enable_mapping given, setting to default value of false")
 		} else {
-			optionalConfigParams.EnableMapping = config.CloudStoryEnabled
+			optionalConfigParams.EnableMapping = *config.EnableMapping
+		}
+
+		if err := validateModes(optionalConfigParams, config.IMUIntegrationEnabled); err != nil {
+			return OptionalConfigParams{}, err
 		}
 		return optionalConfigParams, nil
 	}
@@ -178,4 +186,19 @@ func GetOptionalParameters(config *Config, defaultLidarDataRateMsec, defaultIMUD
 	}
 
 	return optionalConfigParams, nil
+}
+
+func validateModes(
+	optionalConfigParams OptionalConfigParams,
+	imuEnabled bool,
+) error {
+	if optionalConfigParams.LidarDataRateMsec == 0 &&
+		!optionalConfigParams.EnableMapping &&
+		optionalConfigParams.ExistingMap != "" {
+		if imuEnabled {
+			return errLocalizationInOfflineModeIMU
+		}
+		return errLocalizationInOfflineMode
+	}
+	return nil
 }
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -161,8 +165,13 @@ func GetOptionalParameters(config *Config, defaultLidarDataRateMsec, defaultIMUD
 		if config.EnableMapping == nil {
 			logger.Debug("no enable_mapping given, setting to default value of false")
 		} else {
-			optionalConfigParams.EnableMapping = config.CloudStoryEnabled
+			optionalConfigParams.EnableMapping = *config.EnableMapping
```
kim-mishra, line 168: this was a bug

config/config.go
```diff
@@ -179,3 +188,18 @@ func GetOptionalParameters(config *Config, defaultLidarDataRateMsec, defaultIMUD
 
 	return optionalConfigParams, nil
 }
+
+func validateModes(
+	optionalConfigParams OptionalConfigParams,
+	imuEnabled bool,
+) error {
+	localizationMode := optionalConfigParams.LidarDataRateMsec == 0
```
EshaMaharishi, line 196: This seems wrong - isn't it that if LidarDataRateMsec == 0, we're in *offline* mode, not *localization* mode?

config/config.go
```diff
@@ -179,3 +188,18 @@ func GetOptionalParameters(config *Config, defaultLidarDataRateMsec, defaultIMUD
...
+	offlineMode := !optionalConfigParams.EnableMapping && optionalConfigParams.ExistingMap != ""
```
EshaMaharishi, line 197: Similarly here, should this be setting localizationMode?

config/config.go
kim-mishra: yes got them backwards fixing!

config/config.go
kim-mishra: yes got them backwards fixing!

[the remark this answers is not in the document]
kim-mishra: oh you are right - removing `existing_map` check

config/config.go
```diff
@@ -43,8 +44,11 @@ type OptionalConfigParams struct {
 }
 
 var (
-	errCameraMustHaveName    = errors.New("\"camera[name]\" is required")
-	errSensorsMustNotBeEmpty = errors.New("\"sensors\" must not be empty")
+	errCameraMustHaveName              = errors.New("\"camera[name]\" is required")
+	errSensorsMustNotBeEmpty           = errors.New("\"sensors\" must not be empty")
+	errLocalizationInOfflineModeSuffix = "= 0, enable_mapping = false, and existing_map provided. localization in offline mode not supported."
```
EshaMaharishi, line 49: should this message be updated to not mention existing_map?

config/config.go
```diff
@@ -43,8 +44,11 @@ type OptionalConfigParams struct {
...
+	errLocalizationInOfflineMode       = newError(fmt.Sprintf("data_rate_msec %s", errLocalizationInOfflineModeSuffix))
+	errLocalizationInOfflineModeIMU    = newError(fmt.Sprintf("camera[data_freq_hz] %s", errLocalizationInOfflineModeSuffix))
```
EshaMaharishi, line 51: [optional] if we do need two messages, suggestion to just write the full message out twice for easier readability

config/config.go
```diff
@@ -179,3 +188,18 @@ func GetOptionalParameters(config *Config, defaultLidarDataRateMsec, defaultIMUD
...
+	optionalConfigParams OptionalConfigParams,
+	imuEnabled bool,
+) error {
+	offlineMode := optionalConfigParams.LidarDataRateMsec == 0
+	localizationMode := !optionalConfigParams.EnableMapping
+	if localizationMode && offlineMode {
+		if imuEnabled {
+			return errLocalizationInOfflineModeIMU
+		}
+		return errLocalizationInOfflineMode
```
EshaMaharishi, line 202: i'm also confused why we return errLocalizationInOfflineMode that mentions data_rate_msec in its message when we aren't checking data_rate_msec at all?

config/config_test.go
```diff
@@ -297,6 +297,37 @@ func getOptionalParametersTestHelper(
 		}
 	})
 
+	t.Run(fmt.Sprintf("config that puts cartographer in offline mode and in localization mode %s", suffix), func(t *testing.T) {
+		cfgService := makeCfgService(imuIntegrationEnabled, cloudStoryEnabled)
+		cfgService.Attributes["existing_map"] = "test-file.pbstream"
+		cfgService.Attributes["data_rate_msec"] = 0
```
EshaMaharishi, line 303: why does the config have both data_rate_msec and camera.data_frequency_hz? shouldn't it have one or the other?

kim-mishra: `data_rate_msec` gets ignored if IMU is enabled, but I can only add it to the config if imu is enabled if you think that will make it clearer

config/config.go
kim-mishra: This caused a lint error for too many characters in the line

config/config.go
kim-mishra: discussed offline

config/config_test.go
kim-mishra: made this change!

config/config.go
EshaMaharishi: when IMU is not enabled, `optionalConfigParams.LidarDataRateMsec` comes from `data_rate_msec`

EshaMaharishi (approved):
Lgtm!

PR #252, opened 2023-08-21
[RSDK-4448] Flaky test - TestCartographerIntegration2D

PStrutz:
Hot fix for integration test

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -146,7 +146,7 @@ func testHelperCartographer(
 		attrCfg.MovementSensor = map[string]string{"name": "stub_imu"}
 	}
 
-	done := make(chan struct{})
+	lidarDone := make(chan struct{})
 	imuDone := make(chan struct{})
 	sensorReadingInterval := time.Millisecond * 200
 	timedLidar, err := testhelper.IntegrationTimedLidarSensor(t, attrCfg.Camera["name"], replaySensor, sensorReadingInterval, lidarDone)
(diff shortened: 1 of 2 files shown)


integration_test.go
```diff
@@ -134,10 +147,11 @@ func testHelperCartographer(
 	}
 
 	done := make(chan struct{})
```
kim-mishra, line 149:
```suggestion
	lidarDone := make(chan struct{})
```

integration_test.go
```diff
@@ -134,10 +147,11 @@ func testHelperCartographer(
...
+	imuDone := make(chan struct{})
 	sensorReadingInterval := time.Millisecond * 200
 	timedLidar, err := testhelper.IntegrationTimedLidarSensor(t, attrCfg.Camera["name"], replaySensor, sensorReadingInterval, done)
```
kim-mishra, line 152:
```suggestion
	timedLidar, err := testhelper.IntegrationTimedLidarSensor(t, attrCfg.Camera["name"], replaySensor, sensorReadingInterval, lidarDone)
```

integration_test.go
```diff
@@ -157,16 +171,17 @@ func testHelperCartographer(
 	if !utils.SelectContextOrWaitChan(ctx, done) {
 		test.That(t, errors.New("test timeout"), test.ShouldBeNil)
 	}
```
kim-mishra, line 173: Should we wait for `imuDone` here as well?

PStrutz: No, since we are currently never passing done to the imuDone channel

kim-mishra: cool this is a nit but can we move the comment above the `selectContextOrWaitChan`. I think comments typically refer to what is below it

kim-mishra (approved):
LGTM, I think it would be useful to add a comment somewhere that says `IntegrationTimedIMUSensor` will not be used for now also because I would have expected it to have gone through the IMU readings listed out. thanks for the fix!

PR #253, opened 2023-08-22
[RSDK-4659] Divide by 0 error when camera and movement_sensor frequency are 0

PStrutz:


config/config.go
```diff
@@ -132,13 +132,23 @@ func GetOptionalParameters(config *Config, defaultLidarDataRateMsec, defaultIMUD
 			optionalConfigParams.ImuName = imuName
 			strMovementSensorDataFreqHz, ok := config.MovementSensor["data_frequency_hz"]
 			if !ok {
```
EshaMaharishi, line 134: mind changing `ok` to `exists` while you're here?

config/config.go
```diff
@@ -132,13 +132,23 @@ func GetOptionalParameters(config *Config, defaultLidarDataRateMsec, defaultIMUD
...
-				logger.Debugf("config did not provide movement_sensor[data_frequency_hz], setting to default value of %d", 1000/defaultIMUDataRateMsec)
+				if optionalConfigParams.LidarDataRateMsec == 0 {
+					optionalConfigParams.ImuDataRateMsec = 0
+					logger.Warn("camera[data_frequency_hz] is set to 0, " +
+						"setting movement_sensor[data_frequency_hz] to 0")
+				} else {
+					optionalConfigParams.ImuDataRateMsec = defaultIMUDataRateMsec
```
EshaMaharishi, line 140: Should this be `1000 / defaultIMUDataRateMsec`?

EshaMaharishi (approved):
Lgtm mod comments!

config/config.go
PStrutz: will do

kim-mishra (approved):
LGTM except the mismatch between code and log on ins 140-143, which Esha already pointed out. Although I though the code was correct and the log incorrect

optionalConfigParams.ImuDataRateMsec = defaultIMUDataRateMsec
logger.Warn("config did not provide movement_sensor[data_frequency_hz], "+
						"setting to default value of %d", 1000/defaultIMUDataRateMsec)

config/config.go
PStrutz: No - we are passing the default in as a data rate, so no conversion needs to happen

kim-mishra: can we fix the log on the next line in that case

PR #254, opened 2023-08-23
RSDK-4027 - update doCommand to report when cartographer has processed all data

kim-mishra:
# Changes
* add capability to call `RunFinalOptimization` through the CAPI queue and relevant tests along the way
*  call `RunFinalOptimization` before reporting that cartographer has finished


viam-cartographer/src/carto_facade/carto_facade.cc
```diff
@@ -1200,3 +1212,22 @@ extern int viam_carto_get_internal_state_response_destroy(
...
+        cf->RunFinalOptimization();
+    } catch (int err) {
+        return err;
+    } catch (std::exception &e) {
+        LOG(ERROR) << e.what();
+        return VIAM_CARTO_UNKNOWN_ERROR;
+    }
+
+    return VIAM_CARTO_SUCCESS;
+};
```
JohnN193, line 1233: newline

sensorprocess/sensorprocess.go
```diff
@@ -37,6 +38,10 @@ func Start(
 			return false
 		default:
 			if jobDone := addLidarReading(ctx, config); jobDone {
+				err := config.RunFinalOptimizationFunc(ctx, config.Timeout)
```
JohnN193, line 41: is there anything we need to add for other sensors? like if the lidar completes its own offline dataset before the imu, do we need to block the imu from trying to add sensors? @jeremyrhyde

JohnN193 (commented):
going to test this in a CRJ first, will approve when thats done!

JohnN193 (approved):
LGTM, offline mode was successful in the CRJ. Cartographer also posts an info log when the optimization completes, which is a nice surprise. 

would still want some info to make sure this works fine with IMUs in offline mode. but integration tests are passing so maybe its okay?

sensorprocess/sensorprocess.go
jeremyrhyde: I will add the additional blocking mechanisms i nthe IMU PR. Should be a simple:

```
if jobDone := addLidarReading(ctx, config); jobDone {
        stopped = true
        err := config.RunFinalOptimizationFunc(ctx, config.Timeout)
```

PR #242, opened 2023-08-08
[RSDK-4110] Feed IMU data to cartographer

PStrutz:
Create second sensorprocess thread for IMU
Create addIMUReading function in cartofacade and C code
Create CAPI wrapper for addIMUReading

TESTING
*note that this PR will not yet contain the offline tests, as there is a different ticket for creating the replay IMU stub dependencies, upon which the tests are dependent

PR #255, opened 2023-08-25
RSDK-4110 IMU Data Integration

jeremyrhyde:
**Summary**
This PR introduces IMU support to viam's implementation of cartographer. The additional cgo interface functions have been created and linked to the imu data sensor ingestion functions in cartographer. Beginning by matching the pattern of lidar, this PR also includes a more complex sensor processing setup to allow for multiple sensor pipelines to occur in parallel. 

In addition to the new go routine, additional work has been done for offline (replay) datasets due to the need to keep data aligned. This includes:

- Ensure that calls to replay data with the associated methods Linear Acceleration and Angular Velocity return data that is within 20 ms of one another (Note: this tolerance could be made to be dependent on the data_rate_hz down the line)
- Ensure that adding of lidar and imu data occurs sequentially by holding back on pushing more data until all remaining data from the other sensor has been processed
- Allowing the required camera sensor to dictate the beginning and end of the sensor process loops

**Unit Testing**
Thorough unit testing has been done for the imu integration in the cartographer process, sensor process, integration testing and cartofacade functions.

**Offline Testing**
Additional manual testing has be done in offline datasets as seen here:

Without IMU
NewDatasetWithoutIMU

With IMU:
NewDatasetWithIMU

**Live Testing:**

https://github.com/viamrobotics/viam-cartographer/assets/34897732/386a07f3-96e2-4945-8de7-6a8a72ed0c52

**JIRA Ticket:** https://viam.atlassian.net/browse/RSDK-4110


cartofacade/capi_test.go
```diff
@@ -226,14 +246,14 @@ func TestCGoAPI(t *testing.T) {
 		// As a result, these tests show best case behavior.
 
 		// 1. test valid addLidarReading: valid reading ascii
```
kim-mishra, line 248:
```suggestion
		// 1. test valid addSensorReading: valid reading ascii
```

cartofacade/capi_test.go
```diff
@@ -278,14 +298,14 @@ func TestCGoAPI(t *testing.T) {
...
 		timestamp = timestamp.Add(time.Second * 2)
-		testAddSensorReading(t, vc, "viam-cartographer/mock_lidar/2.pcd", timestamp, pointcloud.PCDBinary)
+		testAddLidarReading(t, vc, "viam-cartographer/mock_lidar/2.pcd", timestamp, pointcloud.PCDBinary)
 
 		// test getPosition, is no longer zeroed
 		position, err = vc.getPosition()
 		test.That(t, err, test.ShouldBeNil)
-		test.That(t, position.ComponentReference, test.ShouldEqual, "mysensor")
+		test.That(t, position.ComponentReference, test.ShouldEqual, "mylidar")
 		test.That(t, position.X, test.ShouldNotEqual, 0)
```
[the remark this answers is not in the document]
kim-mishra: @jeremyrhyde what do you mean by the tests suggesting that it is no longer zero?

viam-cartographer/src/carto_facade/carto_facade.h
```diff
@@ -97,20 +108,23 @@ typedef enum viam_carto_LIDAR_CONFIG {
...
 #define VIAM_CARTO_LIB_ALREADY_INITIALIZED 25
-#define VIAM_CARTO_GET_INTERNAL_STATE_RESPONSE_INVLALID 26
+#define VIAM_CARTO_GET_INTERNAL_STATE_RESPONSE_INVALID 26
 #define VIAM_CARTO_GET_INTERNAL_STATE_FILE_WRITE_IO_ERROR 27
 #define VIAM_CARTO_GET_INTERNAL_STATE_FILE_READ_IO_ERROR 28
 #define VIAM_CARTO_NOT_IN_INITIALIZED_STATE 29
 #define VIAM_CARTO_NOT_IN_IO_INITIALIZED_STATE 30
 #define VIAM_CARTO_NOT_IN_STARTED_STATE 31
 #define VIAM_CARTO_NOT_IN_TERMINATABLE_STATE 32
+#define VIAM_CARTO_IMU_ENABLED_INVALID 33
```
kim-mishra, line 125: What does this error mean?

cartofacade/capi_test.go
```diff
@@ -327,3 +347,151 @@ func TestCGoAPI(t *testing.T) {
...
+		// It is not clear why cartographer has this behavior.
+		// Cartographer does not provide any feedback regarding
+		// why or when it does / does not update the map.
+		// As a result, these tests show best case behavior.
+
+		// 1. test valid addLidarReading: valid reading ascii
+		tDelta := time.Second * 5
+		t.Log("lidar reading 1")
+		timestamp = timestamp.Add(tDelta)
+		testAddLidarReading(t, vc, "viam-cartographer/mock_lidar/0.pcd", timestamp, pointcloud.PCDAscii)
```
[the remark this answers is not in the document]
kim-mishra: Do you mean that with IMU data, cartographer needs more data to generate a map?

viam-cartographer/src/carto_facade/carto_facade.h
jeremyrhyde: This error exists for when the `use_imu_data` has been set to true but no imu has been given, or if `use_imu_data` has been set to false and an imu has been given. As `use_imu_data` is set via the presence or absence of an imu, this should never be hit during normal operation.

cartofacade/capi_test.go
jeremyrhyde: discussed offline

sensorprocess/sensorprocess.go
```diff
@@ -115,3 +145,150 @@ func tryAddLidarReading(ctx context.Context, reading []byte, readingTime time.Ti
...
+		default:
+			if config.stopped {
+				return true
+			}
+			if jobDone := config.addIMUReading(ctx); jobDone {
+				return true
+			}
+		}
+	}
+}
```
kim-mishra, line 167: I think it might be a bit cleaner to put this in Start so that we do not have to maintain stopping multiple processes

jeremyrhyde: I looked into doing it in one loop. While offline mode will work well, the issue is live mode. We would need to use tickers vs the sleeps, and the integration tests now produce differing results (this final one may be caused by `VIAM_CARTO_UNABLE_TO_ACQUIRE_LOCK` error, which I would need to look into more):

https://github.com/jeremyrhyde/viam-cartographer/pull/1

I think sticking with the two data loops for now ends up producing a cleaner setup, happy to talk about this more offline if you feel differently

viam-cartographer/src/carto_facade/carto_facade.h
kim-mishra: What do you think about calling it something a little more descriptive of that scenario? I think I would be confused if I got that error. 

Maybe `VIAM_CARTO_IMU_CONFIG_INVALID` or maybe two different errors: `VIAM_CARTO_IMU_ENABLED_BUT_NO_IMU_PROVIDED` & `VIAM_CARTO_IMU_PROVIDED_BUT_IMU_NOT_ENABLED`

sensorprocess/sensorprocess.go
kim-mishra: I see what you mean and agree - sorry for sending you down a rabbit hole!

sensorprocess/sensorprocess.go
```diff
@@ -91,17 +120,18 @@ func tryAddLidarReadingUntilSuccess(ctx context.Context, reading []byte, reading
...
+				config.Logger.Warnw("Retrying sensor reading due to error from cartofacade", "error", err)
 			}
+			config.Logger.Debugf("%v \t | LIDAR | Failure \t \t | %v \n", readingTime, readingTime.Unix())
 		}
 	}
 }
 
-// tryAddLidarReading adds a reading to the carto facade
-// does not retry (online).
+// tryAddLidarReading adds a reading to the carto facade does not retry (online).
```
kim-mishra, line 134:
```suggestion
// tryAddLidarReading adds a reading to the carto facade and does not retry (online).
```

sensorprocess/sensorprocess.go
kim-mishra: Do we throw away IMU readings that occur after the last lidar reading?

sensors/sensors.go
```diff
@@ -211,45 +220,85 @@ func (lidar Lidar) TimedLidarSensorReading(ctx context.Context) (TimedLidarSenso
...
+	var timeLinearAcc, timeAngularVel time.Time
+	var linAcc r3.Vector
+	var angVel spatialmath.AngularVelocity
+	var err error
+	for {
+		select {
+		case <-ctx.Done():
+			return TimedIMUSensorReadingResponse{}, nil
+		default:
+			if timeLinearAcc == defaultTime || timeLinearAcc.Sub(timeAngularVel).Milliseconds() < 0 {
```
kim-mishra, line 246: Could you explain this a bit more?

sensors/test_deps.go
```diff
@@ -198,6 +239,7 @@ func getIMUWithErroringFunctions() *inject.MovementSensor {
 	imu.AngularVelocityFunc = func(ctx context.Context, extra map[string]interface{}) (spatialmath.AngularVelocity, error) {
 		return spatialmath.AngularVelocity{}, errors.New("invalid sensor")
 	}
+
```
kim-mishra, line 242: maybe remove this or add a new line below 
```
imu.LinearAccelerationFunc = func(ctx context.Context, extra map[string]interface{}) (r3.Vector, error) {
		return r3.Vector{}, errors.New("invalid sensor")
	}
```

viam-cartographer/src/carto_facade/carto_facade.cc
```diff
@@ -1201,7 +1300,7 @@ extern int viam_carto_get_internal_state(
 extern int viam_carto_get_internal_state_response_destroy(
     viam_carto_get_internal_state_response *r) {
     if (r == nullptr) {
-        return VIAM_CARTO_GET_INTERNAL_STATE_RESPONSE_INVLALID;
+        return VIAM_CARTO_GET_INTERNAL_STATE_RESPONSE_INVALID;
```
kim-mishra, line 1303: thansk for fixing this!

viam-cartographer/src/carto_facade/carto_facade_test.cc
```diff
@@ -169,41 +209,62 @@ BOOST_AUTO_TEST_CASE(CartoFacade_init_validate) {
...
 
-    BOOST_TEST(viam_carto_init(&vc, lib, vcc, ac) == VIAM_CARTO_SUCCESS);
+    BOOST_TEST(viam_carto_init(&vc, lib, vcc_with_imu_succ, ac) ==
+               VIAM_CARTO_SUCCESS);
     BOOST_TEST(vc->slam_mode == VIAM_CARTO_SLAM_MODE_MAPPING);
 
     BOOST_TEST(viam_carto_terminate(&vc) == VIAM_CARTO_SUCCESS);
-    // can't terminate a carto instance that has already been terminated
-    BOOST_TEST(viam_carto_terminate(&vc) == VIAM_CARTO_VC_INVALID);
+    BOOST_TEST(viam_carto_terminate(&vc) ==
```
kim-mishra, line 234: Why move the comment here?

viam_cartographer.go
```diff
@@ -225,14 +240,18 @@ func New(
...
 	imu := IMU{
 		name:         optionalConfigParams.ImuName,
 		dataRateMsec: optionalConfigParams.ImuDataRateMsec,
 		actual:       imuObject,
-		testing:      timedIMU,
+		timed:        timedIMU,
+	}
+
+	if lidar.dataRateMsec == 0 && imu.dataRateMsec != 0 {
+		logger.Warn("In offline mode, but IMU data frequency is nonzero")
```
kim-mishra, line 254: Should this be an error?

viam_cartographer.go
```diff
@@ -133,11 +137,22 @@ func initSensorProcess(cancelCtx context.Context, cartoSvc *CartographerService)
...
 	}()
+
+	if spConfig.IMUName != "" {
+		cartoSvc.sensorProcessWorkers.Add(1)
+		go func() {
+			defer cartoSvc.sensorProcessWorkers.Done()
+			if jobDone := spConfig.StartIMU(cancelCtx); jobDone {
+				cartoSvc.jobDone.Store(true)
+				cartoSvc.cancelSensorProcessFunc()
+			}
```
kim-mishra, line 153: I think this might cause an issue because if the replay IMU returns an `errEndOfDataset` we will cancel the lidar sensor process before final optimization is done

kim-mishra (commented):
have some clarifying questions (mostly about the sensor process) and some nits

sensorprocess/sensorprocess.go
jeremyrhyde: We want to stop adding imu data after lidar data has been complete for two reasons:
1. For optimization runs after lidar data has been complete https://github.com/viamrobotics/viam-cartographer/pull/254#discussion_r1304608372
2. Once we are no longer at getting lidar data, our constraint equations for loop closure would only be relying on imu data which may lead to errors. The IMU constraint equations are supposed to provide additional data, not be the sole arbitrator of pose estimation.

sensorprocess/sensorprocess.go
```diff
@@ -50,35 +68,46 @@ func Start(
...
+		// sleep remainder of time interval
+		timeToSleep := tryAddLidarReading(ctx, tsr.Reading, tsr.ReadingTime, *config)
 		time.Sleep(time.Duration(timeToSleep) * time.Millisecond)
 		config.Logger.Debugf("sleep for %s milliseconds", time.Duration(timeToSleep))
+	} else {
+		// add the stored lidar data and begin processing the next one, if no imu exists or the currently stored imu data occurs after it.
+		if config.IMUName == "" || config.nextData.lidarTime.Sub(config.nextData.imuTime).Milliseconds() <= 0 {
+			if config.nextData.lidarData != nil {
+				tryAddLidarReadingUntilSuccess(ctx, config.nextData.lidarData, config.nextData.lidarTime, *config)
+				config.started = true
```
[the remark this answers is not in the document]
jeremyrhyde: What `started` does, is only allows imu data to be sent to cartographer after the first lidar data is added. If started is not set, it will continue to poll for more imu data until one is found after the first lidar reading. This allows that lidar reading to be added, `started` to be set to true, and imu data to begin being added.

Similar to stopped, we only want to add imu data when there is lidar data available for the constraint equations to work off of. Without this, I experienced issues where saved replay data was gathered from the movement sensors several seconds in advance of lidar data leading to ~60 imu measurements being added to carto, (and failing) before the first lidar measurement could occur.

sensorprocess/sensorprocess.go
```diff
@@ -91,17 +120,18 @@ func tryAddLidarReadingUntilSuccess(ctx context.Context, reading []byte, reading
 		default:
 			err := config.CartoFacade.AddLidarReading(ctx, config.Timeout, config.LidarName, reading, readingTime)
 			if err == nil {
+				config.Logger.Debugf("%v \t | LIDAR | Success \t \t | %v \n", readingTime, readingTime.Unix())
 				return
 			}
 			if !errors.Is(err, cartofacade.ErrUnableToAcquireLock) {
-				config.Logger.Warnw("Skipping sensor reading due to error from cartofacade", "error", err)
+				config.Logger.Warnw("Retrying sensor reading due to error from cartofacade", "error", err)
```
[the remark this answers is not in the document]
jeremyrhyde: No, it is retrying as this is in `tryAddLidarReadingUntilSuccess`. In offline mode, if we aren't able to acquire a lock, we want to keep trying until we do, as we dont care about being in real-time and needing to process the most up-to-date measurement.

sensors/sensors.go
jeremyrhyde: Similar to the imu and lidar sensor process what we are doing here is trying to align data. However, in this case, we are creating a single packet and therefore need to search for linear acceleration and angular velocity data whose timestamps are within a certain tolerance of one another (currently defined to be 20 ms).

In order to do this we keep track of the timestamps of the current measurements and re-poll for whichever measurement is before the other, until we have two values within this tolerance.

jeremyrhyde: Note: currently I am using a magic number of 20ms for this tolerance. What are your thoughts on me using a value correlated to the imu_data_rate. Say `tolerance = imu_data_rate` or tolerance = `imu_data_rate/2`?

sensors/test_deps.go
jeremyrhyde: Removed all new lines to match patterns established in lidar functions

viam_cartographer.go
jeremyrhyde: I can see it both ways, either it should error out as it is an incongruity in the config, or we want to warn the user that we will override the imu data rate as the lidar is the primary sensor. What are your thoughts?

viam_cartographer.go
kim-mishra: cool thanks!

viam_cartographer.go
kim-mishra: I think this case should error out and only override in the case that  `lidar.dataRateMsec == 0` and `imu.dataRateMsec` is not provided

sensorprocess/sensorprocess.go
```diff
@@ -15,64 +17,91 @@ import (
...
+	IMUDataRateMsec   int
 	Timeout           time.Duration
 	Logger            golog.Logger
+	nextData          nextData
+	started           bool
+}
+
+// nextData stores the next data to be added to cartographer along with its associated timestamp so that,
+// in offline mode, data from multiple sensors can be added in order.
+type nextData struct {
```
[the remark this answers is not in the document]
kim-mishra: Maybe, it might make the checks a bit clearer to the reader

sensorprocess/sensorprocess.go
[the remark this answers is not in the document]
jeremyrhyde: Yes, it was discussed with Nick in Patricia's original PR: https://github.com/viamrobotics/viam-cartographer/pull/242#discussion_r1291547596

sensorprocess/sensorprocess.go
```diff
@@ -10,34 +10,52 @@ import (
...
+	imuData   cartofacade.IMUReading
+}
+
+// StartLidar polls the lidar to get the next sensor reading and adds it to the cartofacade.
 // stops when the context is Done.
-func Start(
+func (config *Config) StartLidar(
 	ctx context.Context,
-	config Config,
 ) bool {
```
kim-mishra, line 51: what do you think of adding this chunk of code here so that we only set started once? Not sure if its worth making this function clunkier but because it happens everytime we add a lidar reading when I first read it I though we were setting this back and forth

```
if config.LidarataRateMsec == 0 {
    tryAddLidarReadingUntilSuccess(ctx, tsr.Reading, tsr.ReadingTime, config)
    config.started = true
}
```

viam-cartographer/src/carto_facade/carto_facade.h
jeremyrhyde: Since this is an uncommon error I'd prefer to keep it as one error. How about `VIAM_CARTO_IMU_PROVIDE_AND_IMU_ENABLED_MISMATCH`?

sensorprocess/sensorprocess.go
jeremyrhyde: Spoke offline, we are moving towards using a start and end time instead of bools to more clearly describe whats happening

kim-mishra (approved):
LGTM, thanks for all of the changes! There are a few things left from the comments like: 
1. splitting `nextData` into `nextIMU` and `nextLidar`
2. changing the hardcoded imu timestamp tolerance to be based on the data rate msec

But Il leave those to you!

sensors/sensors.go
```diff
@@ -15,10 +16,18 @@ import (
 	"go.viam.com/rdk/pointcloud"
 	"go.viam.com/rdk/resource"
 	"go.viam.com/rdk/spatialmath"
+	rdkutils "go.viam.com/rdk/utils"
 	"go.viam.com/rdk/utils/contextutils"
 	goutils "go.viam.com/utils"
 )
 
+const (
+	replayTimeTolerance         = 10 // Milliseconds
```
kkufieta, line 25: [nit] I think the var name should contain the time unit, e.g.:

```suggestion
	replayTimeToleranceMsec         = 10 // Milliseconds
```

(not sure if `Msec` is the convention we follow)

viam-cartographer/src/carto_facade/carto_facade_test.cc
```diff
@@ -1141,14 +1205,467 @@ BOOST_AUTO_TEST_CASE(CartoFacade_start_stop) {
...
+    BOOST_TEST(viam_carto_init(&vc, lib, vcc, ac) == VIAM_CARTO_SUCCESS);
+    BOOST_TEST(vc->slam_mode == VIAM_CARTO_SLAM_MODE_MAPPING);
+    viam::carto_facade::CartoFacade *cf =
+        static_cast<viam::carto_facade::CartoFacade *>(vc->carto_obj);
+    BOOST_TEST(((cf->state) == CartoFacadeState::IO_INITIALIZED));
+
+    BOOST_TEST(fs::is_directory(cf->path_to_internal_state));
+    BOOST_TEST(fs::is_empty(cf->path_to_internal_state));
+
+    // // Start
```
kkufieta, line 1226:
```suggestion
    // Start
```

viam_cartographer.go
```diff
@@ -205,6 +217,10 @@ func New(
 		return nil, err
 	}
 
+	if optionalConfigParams.LidarDataRateMsec == 0 && optionalConfigParams.ImuDataRateMsec != 0 {
```
kkufieta, line 220: [q] What if it is the other way around, i.e. when `LidarDataRateMsec != 0` and `ImuDataRateMsec == 0`?

kkufieta (approved):
LGTM mod questions/nits

viam_cartographer.go
jeremyrhyde: Good point!

PR #256, opened 2023-08-25
increase timeout to 5 minutes

EshaMaharishi:


viam_cartographer.go
```diff
@@ -47,7 +47,7 @@ const (
 	defaultSensorValidationIntervalSec   = 1
 	parsePortMaxTimeoutSec               = 60
 	localhost0                           = "localhost:0"
-	defaultCartoFacadeTimeout            = 5 * time.Second
+	defaultCartoFacadeTimeout            = 5 * 60 * time.Second
```
jeremyrhyde, line 50: Other option
```suggestion
	defaultCartoFacadeTimeout            = 5 * time.Minute
```

EshaMaharishi: good point :) i am not merging this though, it was just so i could use CI to get an appimage

PR #258, opened 2023-08-31
[RSDK-4738] increase timeout for RunFinalOptimization

JohnN193:
ran into an issue where RunFinalOptimization would timeout before completing, which resulted in the C++ code being blocked, even though the though there was no work in the queue for C. This fix increases the timeout to allow for optimization to complete 

ticket https://viam.atlassian.net/browse/RSDK-4738


testhelper/testhelper.go
```diff
@@ -383,6 +384,7 @@ func CreateSLAMService(
 		SensorValidationMaxTimeoutSecForTest,
 		SensorValidationIntervalSecForTest,
 		5*time.Second,
+		5*time.Second,
```
kim-mishra, line 387: should these be `15 * time.Minute`

kim-mishra (approved):
LGTM, one q for test

testhelper/testhelper.go
JohnN193: I felt for tests it was fine to leave it as a shorter timeout, but can increase if you prefer

kim-mishra: that makes sense to me. i am okay with either. maybe we just pass in the default variables so its the same as the real code?

JohnN193: I'll make a set of constants for the tests to be consistent with the other constants in testhelper. otherwise I would have to export the defaults in the viam-cartographer package

PR #264, opened 2023-09-11
RSDK-4664 - Return grpc errors for Position until cartographer has set an initial position

kkufieta:
Ticket: https://viam.atlassian.net/browse/RSDK-4664

Done:
* Return gRPC error for Position until cartographer has set an initial position
    * The initial position is set after a second measurement is added. 
    * It is always set to [0, 0, 0] at that point and only changed once the 3rd measurement has been added.


viam-cartographer/src/carto_facade/carto_facade.h
```diff
@@ -441,50 +442,8 @@ class CartoFacade {
...
-    // terminate the program.
-    // void CacheMapInLocalizationMode();
-
-    // std::vector<std::string> file_list_offline;
-    // size_t current_file_offline = 0;
-    // std::string current_file_online;
-
     std::shared_mutex optimization_shared_mutex;
 
-    // std::atomic<bool> finished_processing_offline{false};
```
kkufieta, line 487: Found these left-over comments and confirmed that they're no longer needed and can be deleted.

jeremyrhyde: Good catch!

viam-cartographer/src/carto_facade/carto_facade_test.cc
```diff
@@ -791,19 +791,9 @@ BOOST_AUTO_TEST_CASE(CartoFacade_demo_without_imu) {
...
-        BOOST_TEST(pr.z == 0);
-        BOOST_TEST(pr.imag == 0);
-        BOOST_TEST(pr.jmag == 0);
-        BOOST_TEST(pr.kmag == 0);
-        BOOST_TEST(pr.real == 1);
-        BOOST_TEST(to_std_string(pr.component_reference) == "lidar");
-
-        BOOST_TEST(viam_carto_get_position_response_destroy(&pr) ==
-                   VIAM_CARTO_SUCCESS);
+        // it should return an error
```
jeremyrhyde, line 794:
```suggestion
        // it should return the position uninitialized error
```

jeremyrhyde (commented):
Code looks good, however, I notice that the code will set the local_pose_initialized to true after the first measurement is added. Your description in the PR states that initialization of the pose happens after the second or third measurements.

kkufieta:
> Code looks good, however, I notice that the code will set the local_pose_initialized to true after the first measurement is added. Your description in the PR states that initialization of the pose happens after the second or third measurements.

@jeremyrhyde `local_pose_initialized` is set to true once `GetLocalSlamResultCallback` is called, which happens after the second measurement.

kim-mishra (approved):

PR #266, opened 2023-09-18
RSDK-3400 Separate Build Directories

jeremyrhyde:
Separate viam-cartographer and cartographer builds into separate folders to allow parallel building in macOS and canon. This was tested on aarch64 + amd64 + macOSX, simultaneously.

**JIRA Ticket:** https://viam.atlassian.net/browse/RSDK-3400


Otterverse (approved):
The "clean" target in the makefile should probably only remove the CURRENT architecture outputs, rather than all of build and bin. Otherwise, LGTM.

PR #276, opened 2023-09-27
RSDK-4610 Create New Config Flag

jeremyrhyde:
This PR replaces the "imu_integration_flag" with a "new_config_flag" which is to be used to allow devs to access unreleased config changes (such as imu intergation, cloud story, and odometry) before full migration occurs. Placing these all under one flag reduces the number of breaking changes needed and provides a better control mechanism for timing releases with doc updates. 

**JIRA Ticket:** https://viam.atlassian.net/browse/RSDK-4610


kim-mishra (approved):

JohnN193 (approved):
LGTM, but can we also setup a pr in app to migrate to this for cloudslam?

PR #279, opened 2023-09-29
[RSDK-3906] validation passes when a replay camera does not have data present

JohnN193:
Updates cartographer to pass validation if a replay camera is waiting for data


kim-mishra (approved):
LGTM! it would be helpful to update the comment to fully explain why we break instead of returning here

PR #285, opened 2023-10-09
RSDK-4933 Remove logging for IMU if not configured

vijayvuyyuru:
Hey yall, this PR is to remove logging about a missing movement sensor if it is not configured as it is not actually an error to surface to the user. 
I also changed one log from a warning to debug for the same reasons. 
**Testing:**
Ran `viam-cartographer` locally after changes to make sure no warning logs showed up.


kim-mishra (approved):
LGTM!

PR #286, opened 2023-10-10
RSDK-4742 Release New SLAM Config

jeremyrhyde:
This PR removes the enable_cloud_story and new_config_flag flags and makes them the default, deprecating the old configurations in the process.

**JIRA Ticket:** https://viam.atlassian.net/browse/RSDK-4742


cartofacade/testhelpers.go
```diff
@@ -19,14 +19,13 @@ func GetTestConfig(cameraName, movementSensorName string) (CartoConfig, string,
 		ComponentReference: "component",
 		LidarConfig:        TwoD,
 		CloudStoryEnabled:  false,
-		EnableMapping:      false,
+		EnableMapping:      true,
```
jeremyrhyde, line 22: [note] setting to true so we dont have to provide a apriori map

cartofacade/testhelpers.go
```diff
@@ -19,14 +19,13 @@ func GetTestConfig(cameraName, movementSensorName string) (CartoConfig, string,
...
 	}, dir, nil
 }
 
 // GetBadTestConfig gets a sample config for testing purposes that will cause a failure.
 func GetBadTestConfig() CartoConfig {
 	return CartoConfig{
-		Camera:      "rplidar",
```
jeremyrhyde, line 29: [note] removing as "data_dir" can't be used as a required param anymore, must be camera

config/config.go
```diff
@@ -20,86 +20,51 @@ type Config struct {
...
 
 // OptionalConfigParams holds the optional config parameters of SLAM.
 type OptionalConfigParams struct {
-	LidarDataRateMsec int
-	ImuName           string
-	ImuDataRateMsec   int
-	MapRateSec        int
-	EnableMapping     bool
-	ExistingMap       string
+	LidarDataFrequencyHz int
```
jeremyrhyde, line 31: [note] switched to frequency hz throughout due to confusion between all the difference data rate secs during removal. Can revert this change if people would prefer DataRateMsec

integration_test.go
```diff
@@ -121,16 +120,19 @@ func testCartographerMap(t *testing.T, svc slam.Service, localizationMode bool)
 }
 
 // Saves cartographer's internal state in the data directory.
-func saveInternalState(t *testing.T, internalState []byte, dataDir string) {
+func saveInternalState(t *testing.T, internalState []byte, dataDir string) string {
```
jeremyrhyde, line 123: [note] now returning filename/path since we need to set it in existing_map

integration_test.go
```diff
@@ -285,35 +280,21 @@ func TestIntegrationCartographer(t *testing.T) {
...
-		// Live + imu sensor
+		// Replay with imu sensor
 		{
 			description: "replay with imu sensor mapping mode 2D",
 			replay:      true,
 			imuEnabled:  true,
 			mode:        cartofacade.MappingMode,
 			subAlgo:     viamcartographer.Dim2d,
 		},
-		{
```
jeremyrhyde, line 310: [note] Had to remove replay localizing tests due to error:

` 'SLAM Service configuration error: camera[data_freq_hz] and enable_mapping = false. localization in offline mode not supported.'`

viam_cartographer_test.go
```diff
@@ -31,337 +31,22 @@ const (
...
-
-		cmd := map[string]interface{}{}
-		resp, err := svc.DoCommand(ctx, cmd)
-		test.That(t, resp, test.ShouldBeNil)
-		test.That(t, err, test.ShouldBeError, viamcartographer.ErrUseCloudSlamEnabled)
-
-		test.That(t, svc.Close(context.Background()), test.ShouldBeNil)
-	})
-
-	t.Run("Failed creation of cartographer slam service with more than one sensor", func(t *testing.T) {
```
jeremyrhyde, line 97: [note] repeated tests for feature flag (new config) found below

cartofacade/capi.go
```diff
@@ -376,11 +376,18 @@ func getConfig(cfg CartoConfig) (C.viam_carto_config, error) {
...
-	vcc.data_dir = goStringToBstring(cfg.DataDir)
+	// Remove cloud_story_enabled, map_rate_sec, and data_dir from C++ code
+	// JIRA Ticket: RSDK-52334 https://viam.atlassian.net/browse/RSDK-5334
+	vcc.cloud_story_enabled = C.bool(true)
+	vcc.data_dir = goStringToBstring("/tmp/")
+	if cfg.EnableMapping {
+		vcc.map_rate_sec = C.int(9000)
+	} else {
+		vcc.map_rate_sec = C.int(0)
+	}
```
kim-mishra, line 387: Could you explain what this is trying to do?

kim-mishra, line 387: Could you explain what this will do?

integration_test.go
JohnN193: Not too sure whats going on with this error tbh, unless all of our replay camera tests are assumed to be offline mode only in our tests? 

if this is the correct behavior, should we include tests to make sure we hit this error?

cartofacade/testhelpers.go
kim-mishra: Does this mean that localizing/udpating are no longer being tested?

integration_test.go
```diff
@@ -121,29 +120,32 @@ func testCartographerMap(t *testing.T, svc slam.Service, localizationMode bool)
...
 	timeStamp := time.Now().UTC()
 	internalStateDir := filepath.Join(dataDir, "internal_state")
 	if err := os.Mkdir(internalStateDir, 0o755); err != nil {
 		t.Error("TEST FAILED failed to create test internal state directory")
+		return ""
 	}
 	filename := filepath.Join(internalStateDir, "map_data_"+timeStamp.UTC().Format(testhelper.SlamTimeFormat)+".pbstream")
 	if err := os.WriteFile(filename, internalState, 0o644); err != nil {
 		t.Error("TEST FAILED failed to write test internal state")
+		return ""
```
kim-mishra, line 133:
```suggestion
err := os.WriteFile(filename, internalState, 0o644)
test.That(t, err, test.ShouldBeNil)
```

integration_test.go
```diff
@@ -121,29 +120,32 @@ func testCartographerMap(t *testing.T, svc slam.Service, localizationMode bool)
 }
 
 // Saves cartographer's internal state in the data directory.
-func saveInternalState(t *testing.T, internalState []byte, dataDir string) {
+func saveInternalState(t *testing.T, internalState []byte, dataDir string) string {
 	timeStamp := time.Now().UTC()
 	internalStateDir := filepath.Join(dataDir, "internal_state")
 	if err := os.Mkdir(internalStateDir, 0o755); err != nil {
 		t.Error("TEST FAILED failed to create test internal state directory")
+		return ""
```
kim-mishra, line 128:
```suggestion
err := os.Mkdir(internalStateDir, 0o755)
test.That(t, err, test.ShouldBeNil)
```

kim-mishra (commented):
Mostly LGTM just a couple of questions!

integration_test.go
kim-mishra: I think we should only be in offline mode if`camera[data_freq_hz] = 0` but I would expect it to be the default value here

cartofacade/capi.go
jeremyrhyde: This is an artifact of avoiding the removal of map_rate_sec and data_Dir from the C++ code. Since it still relies on a configured map_rate_sec, I am creating said parameter at the last possible moment based on its new definition.

integration_test.go
jeremyrhyde: Should replay sensors have a non-zero data_freq_hz?

integration_test.go
jeremyrhyde: Great call out!

integration_test.go
JohnN193: for live mode yes, offline mode they should be set to 0

cartofacade/testhelpers.go
jeremyrhyde: Resolved, looks like it was an artifact of an issue I corrected. Reverting back to original `EnableMapping: false`

integration_test.go
jeremyrhyde: Done

integration_test.go
```diff
@@ -64,20 +63,20 @@ func testCartographerPosition(t *testing.T, svc slam.Service, useIMU bool, expec
...
 
 	case runtime.GOOS == "darwin" && useIMU:
-		expectedPos = r3.Vector{X: 4.4700878707562035, Y: 3.1781587655776358, Z: 0}
+		expectedPos = r3.Vector{X: 3.2858556935949954, Y: 2.707920249449911, Z: 0}
 		expectedOri = &spatialmath.R4AA{
-			RX:    0.9861776038047263,
-			RY:    0.1637212678758259,
-			RZ:    0.025477052402116784,
-			Theta: 0.02399255141454847,
+			RX:    0.985674239521596,
```
jeremyrhyde, line 68: [note] These values got updated when I changed the timing of the IMU by 1 millisecond. This change was necessary to support online testing with IMU (before we were doing offline but to support localization mode we had to mvoe to online mode).

integration_test.go
```diff
@@ -250,72 +244,74 @@ func TestIntegrationCartographer(t *testing.T) {
 
 	cases := []struct {
 		description string
+		online      bool
```
jeremyrhyde, line 247: added online versus relay distinction

kim-mishra (approved):
lgtm

config/config.go
tessavitabile: I wouldn't have included this change as part of this PR, since this PR to turn on cloud SLAM should be as low risk and easy to review as possible, to reduce the chance it introduces a bug. However, since you've already included it, I don't think you need to change it.

tessavitabile (approved):
Code changes LGTM! I didn't review the tests. Will someone on the team be able to manually test this with https://github.com/viamrobotics/app/pull/2647 before Tuesday?

JohnN193 (approved):

cartofacade/capi.go
```diff
@@ -376,11 +376,18 @@ func getConfig(cfg CartoConfig) (C.viam_carto_config, error) {
...
 	}
 
-	vcc.map_rate_sec = C.int(cfg.MapRateSecond)
-	vcc.data_dir = goStringToBstring(cfg.DataDir)
+	// Remove cloud_story_enabled, map_rate_sec, and data_dir from C++ code
+	// JIRA Ticket: RSDK-52334 https://viam.atlassian.net/browse/RSDK-5334
+	vcc.cloud_story_enabled = C.bool(true)
+	vcc.data_dir = goStringToBstring("/tmp/")
+	if cfg.EnableMapping {
+		vcc.map_rate_sec = C.int(9000)
```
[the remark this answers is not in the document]
jeremyrhyde: Correct! I've added a comment to that affect

PR #288, opened 2023-10-13
RSDK-5366 - treat empty string for movement sensor name as not providing a moveme…

kim-mishra:
This PR checks if the imuName was provided in the places where we initially only checked if the movement sensor was provided.


tessavitabile:
@jeremyrhyde can you please prioritize this review, since it is a release blocker for cloud slam?

PR #293, opened 2023-11-06
RSDK-4860 - Sensor Support for odometry data (sensors.go)

kkufieta:
Ticket: https://viam.atlassian.net/browse/RSDK-4860


cartofacade/capi.go
```diff
@@ -118,7 +118,6 @@ type CartoAlgoConfig struct {
 	MaxRange             float32
 	MinRange             float32
 	UseIMUData           bool
-	UseOdometerData      bool
```
kkufieta, line 121: Removed this because there is no such configuration parameter in Cartographer. In order to use odometry data all we have to do is to add the sensor data to the map builder - no flag has to be enabled.

sensors/odometer.go
```diff
@@ -0,0 +1,183 @@
...
+				readingTimePosition = time.Now().UTC()
+				if timeRequestedMetadata, ok := md[contextutils.TimeRequestedMetadataKey]; ok {
+					replay = true
+					if readingTimePosition, err = time.Parse(time.RFC3339Nano, timeRequestedMetadata[0]); err != nil {
+						return TimedOdometerSensorReadingResponse{}, errors.Wrap(err, replayTimestampErrorMessage)
+					}
+				}
+			}
+
+			if readingTimeOrientation == defaultTime || readingTimeOrientation.Sub(readingTimePosition).Milliseconds() < 0 {
```
tessavitabile, line 84: Should we check if the times are within `replayTimeToleranceMsec` before making this call? Otherwise we might keep updating both position and orientation indefinitely.

sensors/odometer.go
```diff
@@ -0,0 +1,183 @@
...
+	replay := false
+
+	var readingTimePosition, readingTimeOrientation time.Time
+	var position *geo.Point
+	var orientation spatialmath.Orientation
+	var err error
+	for {
+		select {
+		case <-ctx.Done():
+			return TimedOdometerSensorReadingResponse{}, nil
```
tessavitabile, line 67: I think we should return an error here, if there is one on the context.

sensors/odometer.go
```diff
@@ -0,0 +1,183 @@
...
+
+				readingTimeOrientation = time.Now().UTC()
+				if timeRequestedMetadata, ok := md[contextutils.TimeRequestedMetadataKey]; ok {
+					replay = true
+					if readingTimeOrientation, err = time.Parse(time.RFC3339Nano, timeRequestedMetadata[0]); err != nil {
+						return TimedOdometerSensorReadingResponse{}, errors.Wrap(err, replayTimestampErrorMessage)
+					}
+				}
+			}
+			if math.Abs(float64(readingTimeOrientation.Sub(readingTimePosition).Milliseconds())) < replayTimeToleranceMsec {
```
tessavitabile, line 98: Can you make sure `TimedOdometerSensorReading` is always called with a timeout, in case this condition is never satisfied?

sensors/odometer.go
```diff
@@ -0,0 +1,183 @@
...
+		dataFrequencyHz: dataFrequencyHz,
+		Odometer:        movementSensor,
+	}, nil
+}
+
+// ValidateGetOdometerData checks every sensorValidationIntervalSec if the provided Odometer
+// returned valid timed readings every sensorValidationIntervalSec
+// until either success or sensorValidationMaxTimeoutSec has elapsed.
+// returns an error if at least one invalid reading was returned.
+func ValidateGetOdometerData(
```
tessavitabile, line 149: [opt] Why is this function needed? I would expect it to be sufficient to call GetProperties.

viam_cartographer.go
```diff
@@ -231,9 +232,12 @@ func New(
 	}
 
 	// Get the odometer if it is supported
+	var timedOdometer s.TimedOdometerSensor
 	if odometerSupported {
-		// TODO[kat]: Add when working on https://viam.atlassian.net/browse/RSDK-4860
-		logger.Info("odometer is not yet supported")
+		timedOdometer, err = s.NewOdometer(ctx, deps, movementSensorName, optionalConfigParams.MovementSensorDataFrequencyHz, logger)
```
tessavitabile, line 237: [opt] I'm a little concerned about making both an IMU and an odometer out of the same movement sensor. I think this means we are going to have two separate threads that read from the same movement sensor, which may block each other. I think it would be better to have a single thread that reads from the movement sensor and gets both odometry and IMU data.

kkufieta: That's a good point. I thought about this and discussed it with @kim-mishra & @jeremyrhyde . We decided to use only one process for the movement sensor and then split it into IMU & odometer data within [`tryAddIMUReading`](https://github.com/viamrobotics/viam-cartographer/blob/3905758c2a20a80a09dd19bde97441f1fb92448a/sensorprocess/imusensorprocess.go#L132-L146) (which we'd rename to `tryAddMovementSensorReading`)

tessavitabile: Sounds good! Does it still make sense to have separate IMU and Odometer structs?

kkufieta: No, I don't think so. It'll take a bit of refactoring to change this. Working on it now!

sensors/odometer.go
jeremyrhyde: This is true, GetProperties will already make initial calls to the cloud/actual hardware, therefore the only check we need is that the device returns GetProperties that shows the odom endpoints are supported

kkufieta: This is the case only for replay sensors. In the case of live movement sensors, this check will still help us determine whether or not we're able to get the data we need.

jeremyrhyde: Note: Only localization on the robot is using the actual sensors at this point

kkufieta: Should we remove the validation for both the Lidar and MovementSensor then?

I was also wondering if this level of validation is necessary and don't feel strongly about it.

kkufieta: Though I do lean a bit towards removing the validation.

sensors/test_deps.go
```diff
@@ -30,16 +30,19 @@ var (
 	LinAcc = r3.Vector{X: 1, Y: 1, Z: 1}
 	// AngVel is the successful mock angular velocity result used for testing.
 	AngVel = spatialmath.AngularVelocity{X: 1, Y: .5, Z: 0}
```
kkufieta, line 32: Shuffled around some of the variable names & functions to align their order (made it easier for me to read & check the code)

testhelper/integrationtesthelper.go
```diff
@@ -57,7 +57,7 @@ type TimeTracker struct {
 	Mu *sync.Mutex
 }
 
-// IntegrationTimedLidarSensor returns a mock timed lidar sensor
```
kkufieta, line 60: Integration tests will be changed in this ticket: https://viam.atlassian.net/browse/RSDK-4863

viam_cartographer_test.go
```diff
@@ -113,24 +113,6 @@ func TestNew(t *testing.T) {
 		test.That(t, svc.Close(context.Background()), test.ShouldBeNil)
 	})
 
-	t.Run("Successful creation of cartographer slam service with good lidar with movement"+
-		"sensor that does not support IMU nor odometer", func(t *testing.T) {
```
kkufieta, line 117: Removed this, because we're now failing if a movement sensor does not support at least an IMU or odometer.

viam_cartographer.go
```diff
@@ -326,7 +287,11 @@ func checkIfIMUAndOdometerSupported(ctx context.Context, deps resource.Dependenc
 	// A movement sensor used as an odometer must support Position and Orientation.
 	odometerSupported := properties.PositionSupported && properties.OrientationSupported
 
-	return imuSupported, odometerSupported, nil
+	if !imuSupported && !odometerSupported {
+		return s.ErrMovementSensorNeitherIMUNorOdometer
+	}
+
+	return nil
```
kkufieta, line 294: @tessavitabile @jeremyrhyde I changed the implementation such that we're now failing to start up the slam service if the provided movement sensor does not support at least an IMU or odometer. Is that ok?

viam_cartographer_test.go
```diff
@@ -148,22 +130,6 @@ func TestNew(t *testing.T) {
 		test.That(t, svc.Close(context.Background()), test.ShouldBeNil)
 	})
 
-	t.Run("Failed creation of cartographer slam service with invalid lidar sensor "+
```
kkufieta, line 151: Removed this, because we're not validating in `CreateSLAMService` anymore whether or not the lidar or movement sensor is provides the data we expect from it.

sensors/movementsensor.go
```diff
@@ -0,0 +1,324 @@
...
+	"go.opencensus.io/trace"
+	"go.viam.com/rdk/components/movementsensor"
+	"go.viam.com/rdk/resource"
+	"go.viam.com/rdk/spatialmath"
+	rdkutils "go.viam.com/rdk/utils"
+	"go.viam.com/rdk/utils/contextutils"
+)
+
+const (
+	replayTimeToleranceMsec           = 10 // Milliseconds
```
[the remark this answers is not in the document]
tessavitabile: [nit] Also it looks like we check the timestamp difference even when not using a replay sensor, so I don't think `replayTimeToleranceMsec` is the right name. Or should we only be checking the timestamp difference when using a replay sensor?

sensors/movementsensor.go
```diff
@@ -0,0 +1,324 @@
...
+	var err error
+
+	returnReadingIfTimestampsWithinTolerance := func(readingTimePosition, readingTimeOrientation time.Time,
+		position *geo.Point, orientation *spatialmath.Orientation,
+	) (TimedOdometerReadingResponse, bool) {
+		if readingTimeOrientation.Sub(readingTimePosition).Abs().Milliseconds() < replayTimeToleranceMsec {
+			return TimedOdometerReadingResponse{
+				Position:    position,
+				Orientation: *orientation,
+				ReadingTime: readingTimePosition.Add((readingTimePosition.Sub(readingTimeOrientation) / 2).Abs()),
```
tessavitabile, line 228: Same here, this will be incorrect when readingTimePosition is larger.

viam_cartographer.go
```diff
@@ -214,58 +207,47 @@ func New(
...
 	} else {
-		imuSupported, odometerSupported, err = checkIfIMUAndOdometerSupported(ctx, deps, movementSensorName)
-	}
-
-	// Get the IMU if it is supported
-	var timedIMU s.TimedIMUSensor
-	if imuSupported {
-		timedIMU, err = s.NewIMU(ctx, deps, movementSensorName, optionalConfigParams.MovementSensorDataFrequencyHz, logger)
-		if err != nil {
+		if err = checkIfIMUAndOdometerSupported(ctx, deps, movementSensorName); err != nil {
```
tessavitabile, line 214: [nit] I think this check is redundant with the check in NewMovementSensor() and can be removed.

viam_cartographer.go
tessavitabile: Yes

tessavitabile (approved):
LGTM mod comments. Due to the behavior changes in the loops for getting movement sensor data, I think this change needs to be manually tested to check that we have not introduced regressions in these cases:
1) online mapping in the cloud with IMU
2) offline mapping in the cloud with IMU (though this already may be broken: https://viam.atlassian.net/browse/RSDK-5522)
3) localizing with IMU (though this already may not be performing well: https://viam.atlassian.net/browse/RSDK-5417)

sensors/movementsensor.go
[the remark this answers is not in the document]
jeremyrhyde: The need for this tolerance is that capturing both the LinearAcceleration and AngularVelocity data occurs independently of one another. So there is a chance for this values to be gotten, 'off cycle'.

Ex. data frequency hz : 1 leads to LinearAcceleration being captured at 0, 1, 2, 3... but could lead to AngularVelocity being captured at 0.4, 1.4, 2.4, 3.4...

Ideally, this tolerance should be ~ 1/2 the data frequency or as you say Tess the time difference in the replay sensor. Although, that might be a bit hard to figure out without requesting multiple data points.

sensors/movementsensor.go
```diff
@@ -0,0 +1,324 @@
...
+		linAcc                                      r3.Vector
+		position                                    *geo.Point
+		orientation                                 spatialmath.Orientation
+		timedIMUReadingResponse                     *TimedIMUReadingResponse
+		timedOdometerReadingResponse                *TimedOdometerReadingResponse
+		err                                         error
+	)
+
+	timeoutCtx, cancel := context.WithTimeout(ctx, timedMovementSensorReadingTimeout)
+	defer cancel()
```
jeremyrhyde, line 109: I'm worried about how this section will interact with integration tests. We currently get data from the imu and lidar and compare the times, whichever one one occurs earlier we add in before grabbing new data for the one added. What will happen if imu is before lidar but odom data is after? We may need a way call each one individually.

sensors/movementsensor.go
```diff
@@ -0,0 +1,324 @@
...
+					&readingTimePosition, &readingTimeOrientation); err != nil && !errors.Is(err, ErrNoValidReadingObtained) {
+					return TimedMovementSensorReadingResponse{}, err
+				}
+				if timedOdometerReadingResponse != nil {
+					break odometerLoop
+				}
+			}
+		}
+	}
+	return TimedMovementSensorReadingResponse{
```
jeremyrhyde, line 144: Should we handle the case where imu and odom were requested, odom fails but imu succeeds (and vica versa)?

sensors/movementsensor.go
```diff
@@ -0,0 +1,324 @@
...
+	) (TimedIMUReadingResponse, bool) {
+		if readingTimeAngularVel.Sub(readingTimeLinearAcc).Abs().Milliseconds() < replayTimeToleranceMsec {
+			return TimedIMUReadingResponse{
+				LinearAcceleration: *linAcc,
+				AngularVelocity: spatialmath.AngularVelocity{
+					X: rdkutils.DegToRad(angVel.X),
+					Y: rdkutils.DegToRad(angVel.Y),
+					Z: rdkutils.DegToRad(angVel.Z),
+				},
+				ReadingTime: readingTimeLinearAcc.Add((readingTimeLinearAcc.Sub(readingTimeAngularVel)).Abs() / 2),
```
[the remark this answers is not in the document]
jeremyrhyde: ReadingTime: readingTimeLinearAcc.Add((readingTimeAngularVel.Sub(readingTimeLinearAcc)).Abs()

if readingTimeLinearAcc > readingTimeAngularVel: this becomes the higher - 1/2 the diff
if readingTimeLinearAcc < readingTimeAngularVel: this becomes the lower + 1/2 the diff

sensors/movementsensor.go
```diff
@@ -0,0 +1,324 @@
...
+			}, true
+		}
+		return TimedIMUReadingResponse{}, false
+	}
+
+	if *readingTimeLinearAcc == defaultTime || readingTimeLinearAcc.Sub(*readingTimeAngularVel).Milliseconds() < 0 {
+		ctxWithMetadata, md := contextutils.ContextWithMetadata(ctx)
+		if *linAcc, err = ms.sensor.LinearAcceleration(ctxWithMetadata, make(map[string]interface{})); err != nil {
+			return &TimedIMUReadingResponse{}, errors.Wrap(err, "LinearAcceleration error")
```
jeremyrhyde, line 176: more descriptive error

sensors/movementsensor.go
```diff
@@ -0,0 +1,324 @@
...
+					Y: rdkutils.DegToRad(angVel.Y),
+					Z: rdkutils.DegToRad(angVel.Z),
+				},
+				ReadingTime: readingTimeLinearAcc.Add((readingTimeLinearAcc.Sub(readingTimeAngularVel)).Abs() / 2),
+			}, true
+		}
+		return TimedIMUReadingResponse{}, false
+	}
+
+	if *readingTimeLinearAcc == defaultTime || readingTimeLinearAcc.Sub(*readingTimeAngularVel).Milliseconds() < 0 {
```
jeremyrhyde, line 173: If we use a pointer for readingTimeLinearAcc can this be `readingTimeLinearAcc == nil`?

jeremyrhyde, line 173: Can we turn this section into a for loop over LinearAccleration and Angular velocity instead of repeating the code twice? It would make it more digestible.

sensors/movementsensor.go
```diff
@@ -0,0 +1,324 @@
...
+	}
+
+	if response, ok := returnReadingIfTimestampsWithinTolerance(*readingTimeAngularVel, *readingTimeLinearAcc, angVel, linAcc); ok {
+		return &response, nil
+	}
+
+	return nil, ErrNoValidReadingObtained
+}
+
+func (ms *MovementSensor) timedOdometerReading(ctx context.Context, position *geo.Point, orientation *spatialmath.Orientation,
```
jeremyrhyde, line 216: see previous comments

sensors/movementsensor.go
```diff
@@ -0,0 +1,324 @@
...
+// from a replay sensor.
+type TimedMovementSensor interface {
+	Name() string
+	DataFrequencyHz() int
+	TimedMovementSensorReading(ctx context.Context) (TimedMovementSensorReadingResponse, error)
+	Properties() MovementSensorProperties
+}
+
+// MovementSensorProperties contains information whether or not an IMU and/or odometer are supported.
+type MovementSensorProperties struct {
```
jeremyrhyde, line 46: Since we import movement sensor already, can we just use its properties' definition instead of redefining our own?

viam_cartographer.go
```diff
@@ -123,7 +118,7 @@ func initSensorProcesses(cancelCtx context.Context, cartoSvc *CartographerServic
 	spConfig := sensorprocess.Config{
 		CartoFacade:              cartoSvc.cartofacade,
 		Lidar:                    cartoSvc.lidar,
-		IMU:                      cartoSvc.imu,
+		IMU:                      cartoSvc.movementSensor,
```
jeremyrhyde, line 121: Rename?

viam_cartographer.go
jeremyrhyde: Agreed

viam_cartographer_internal_test.go
```diff
@@ -495,59 +495,29 @@ func TestCheckIfIMUAndOdometerSupported(t *testing.T) {
...
 		lidar, movementSensor := s.NoLidar, s.MovementSensorNotIMUNotOdometer
-		imuSupported, odometerSupported, err := checkIfIMUAndOdometerSupported(ctx,
+		err := checkIfIMUAndOdometerSupported(ctx,
 			s.SetupDeps(lidar, movementSensor), string(movementSensor))
-		test.That(t, err, test.ShouldBeNil)
-		test.That(t, imuSupported, test.ShouldBeFalse)
-		test.That(t, odometerSupported, test.ShouldBeFalse)
-	})
-
-	t.Run("only IMU supported", func(t *testing.T) {
```
jeremyrhyde, line 505: Why were these tests removed?

jeremyrhyde (changes requested):
Looks good, great clean up. I had one larger issue with the inability to get new imu or odom data separately as with replay cameras these may be out of sync and we may need to let one catch up.

sensors/movementsensor.go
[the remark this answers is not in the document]
tessavitabile: 10ms seems too low -- can we make this higher?

And should we either (1) stop checking the timestamp difference when not using a replay sensor, or (2) rename this constant so it doesn't refer to "replay"?

sensors/movementsensor.go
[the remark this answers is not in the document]
tessavitabile: No, if readintTimeLinearAcc > readingTimeAngularVel: this becomes the higher + 1/2 the diff (because of the absolute value)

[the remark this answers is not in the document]
kkufieta: Ah yes, I think I wanted this to be 

```
readingTimeLinearAcc.Add(readingTimeAngularVel).Abs() / 2
```

Do you agree that this is correct?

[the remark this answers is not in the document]
tessavitabile: It's not possible to `Add` two times. I think you can either
1) Convert to millis first, and then average like you showed above. I don't think you will encounter overflow.
2) Check which time is larger, then do either `readingTimeLinearAcc.Add(readingTimeAngularVel.Sub(readingTimeLinearAcc) / 2)` or the reverse

sensors/movementsensor.go
tessavitabile: Oh interesting. I think either we could allow getting IMU data separately from odometry (as you said) or get all of the values together and give them the same timestamp.

kkufieta: This problem will be resolved by changing the way we grab & insert data in offline mode.

sensors/movementsensor.go
kkufieta: This would be resolved if we use Readings in the near future (ticket: https://viam.atlassian.net/browse/RSDK-5736). In that case, getting the reading from the movement sensor would fail as well when one of the endpoints fails. Which is why I think that using the approach I implemented is fine.

sensors/movementsensor.go
kkufieta: What do you mean by "repeating the code twice"?

sensors/movementsensor.go
kkufieta: What are you referring to?

viam_cartographer.go
kkufieta: Not yet, since this is the sensor process which will be handled in another ticket.

viam_cartographer_internal_test.go
kkufieta: Good point. I could still check for these cases!

sensors/movementsensor.go
kkufieta: We can't, since the reading times are already initialized to `time.Time{}` at this point.

sensors/movementsensor.go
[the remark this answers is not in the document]
tessavitabile: I think it would be better to set it based on the data frequency, but I'm fine with this for now.

sensors/movementsensor.go
tessavitabile: Agreed

sensors/movementsensor.go
jeremyrhyde: Agreed, after talking another look I think the added complexity to achieve this wouldn't be beneficial

sensors/movementsensor.go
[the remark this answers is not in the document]
jeremyrhyde: Good catch Tess! I forgot to remove the Abs

jeremyrhyde (approved):
LGTM, spoke offline or one or two last points

kkufieta:
Ran tests and compared performance for the `main` branch vs. this PR. 

The tests ran all on the RPI so I could analyze the logs. The tests I ran were:
* mapping
* replay
* localizing

A visual comparison of the results showed equal performance. No additional error messages were recorded for this PR. The percentage of Lidar measurements vs. IMU measurements that were ingested by cartographer are:

| | `main` branch | This PR |
|--|--|--|
| | Lidar % -- IMU % | Lidar % -- IMU % |
| mapping  | 9.85% -- 90.15%  | 7.86% -- 92.1% |
| replay  | 20.15% -- 79.85%  | 20.1% -- 79.9% |
| localizing  | 20.13% -- 79.86%  | 19.7% -- 80.3% |

The only notable difference is a slight reduction in inserted Lidar measurements in the mapping mode: a drop of 2%. This did not seem to affect the visual results and will be resolved by introducing batch data.

PR #296, opened 2023-11-20
Bump RDK & switch to rdk/logging and rdk/gostream

kkufieta:
Done:
* Bumped RDK to v0.13.0
* Switched over to using `go.viam.com/rdk/logging` instead of `github.com/edaniels/golog`
* Switched over to using `go.viam.com/rdk/gostream` instead of `github.com/viamrobotics/gostream`

I did this while debugging why debug logging did not work (it ended up being an RDK issue), so this is strictly seen as not a necessary change at this moment. Switching over to using `rdk/logging` would be required in future RDK bumps, though, anyway.

Tested on Intel Mac with rplidar in mapping mode.


jeremyrhyde (approved):
LGTM, golog -> logging and gostream update

PR #305, opened 2023-12-08
[RSDK-5694] add viam default parameters to lua files

JohnN193:
https://viam.atlassian.net/browse/RSDK-5694

currently we configure a set of default parameters to use in cartographer within go/c++, but the lua files that we import do not match those defaults. This means if a user were to look at our lua files to see the configuration of cartographer, they would be using a config that does not produce a reasonable map.


kim-mishra (approved):
LGTM

PR #307, opened 2023-12-11
RSDK-4862 - Add odometer sensor process

kkufieta:
Ticket: https://viam.atlassian.net/browse/RSDK-4862

**Tested on:**
* Intel Mac w/ lidar
* Cloud slam (live mode) results show an overall improvement in mapping quality based on this PR. They also show a similar map quality for both IMU and IMU + odometer mapping, though the latter leads to higher confidence sooner than just using an IMU. I could not test mapping using just an odometer because cloud slam won't start unless the data manager is capturing IMU data.

**Cloud slam - live mode:**
NOTE: Unbeknownst to me, the IMU was providing only zero values. The map quality in the results below is therefore not representative of what the actual performance likely looks like. But the maps do still serve as a proof that adding IMU and odometer data is successful.

| | main branch - w/ imu  | this PR - w/ imu   | this PR - w/ imu & odometer | 
| --- | ----------- | ----------- |----------- |
| final result | Screenshot 2023-12-14 at 12 40 51 PM | Screenshot 2023-12-14 at 12 41 22 PM | Screenshot 2023-12-14 at 12 41 39 PM | 
| 1st position | Screenshot 2023-12-14 at 11 46 46 AM | Screenshot 2023-12-14 at 11 56 15 AM | Screenshot 2023-12-14 at 11 28 54 AM | 
| 2nd position | Screenshot 2023-12-14 at 11 47 41 AM | Screenshot 2023-12-14 at 11 56 53 AM | Screenshot 2023-12-14 at 11 30 30 AM | 
| 3rd position | Screenshot 2023-12-14 at 11 48 29 AM | Screenshot 2023-12-14 at 11 57 54 AM | Screenshot 2023-12-14 at 11 31 45 AM | 
| 4th position | Screenshot 2023-12-14 at 11 49 08 AM | Screenshot 2023-12-14 at 11 58 44 AM | Screenshot 2023-12-14 at 11 33 02 AM |

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -38,7 +38,6 @@ type OptionalConfigParams struct {
 
 var (
 	errCameraMustHaveName        = errors.New("\"camera[name]\" is required")
-	errModeHasToBeProvided       = errors.New("\"mode\" config parameter is required")
 	errLocalizationInOfflineMode = newError("\"camera[data_freq_hz]\" and enable_mapping = false. localization in offline mode not supported.")
 )
 
--- a/config/config.go
+++ b/config/config.go
@@ -38,7 +38,7 @@ type OptionalConfigParams struct {
 
 var (
 	errCameraMustHaveName        = errors.New("\"camera[name]\" is required")
-	errLocalizationInOfflineMode = newError("\"camera[data_freq_hz]\" and enable_mapping = false. localization in offline mode not supported.")
+	errLocalizationInOfflineMode = newError("\"camera[data_freq_hz]\" and enable_mapping = false. Localization in offline mode is not supported.")
 )
 
 // Validate creates the list of implicit dependencies.
--- a/config/config.go
+++ b/config/config.go
@@ -38,7 +38,8 @@ type OptionalConfigParams struct {
 
 var (
 	errCameraMustHaveName        = errors.New("\"camera[name]\" is required")
-	errLocalizationInOfflineMode = newError("camera[data_freq_hz] and enable_mapping = false. localization in offline mode not supported.")
+	errModeHasToBeProvided       = errors.New("\"mode\" config parameter is required")
+	errLocalizationInOfflineMode = newError("\"camera[data_freq_hz]\" and enable_mapping = false. localization in offline mode not supported.")
 )
 
 // Validate creates the list of implicit dependencies.
@@ -65,7 +66,8 @@ func (config *Config) Validate(path string) ([]string, error) {
 		deps = append(deps, movementSensorName)
 	}
 
-	if config.ConfigParams["mode"] == "" {
+	mode, ok := config.ConfigParams["mode"]
+	if !ok || mode == "" {
 		return nil, utils.NewConfigValidationFieldRequiredError(path, "config_params[mode]")
 	}
 
--- a/config/config.go
+++ b/config/config.go
@@ -38,7 +38,8 @@ type OptionalConfigParams struct {
 
 var (
 	errCameraMustHaveName        = errors.New("\"camera[name]\" is required")
-	errLocalizationInOfflineMode = newError("\"camera[data_freq_hz]\" and enable_mapping = false. Localization in offline mode is not supported.")
+	errLocalizationInOfflineMode = newError("\"camera[data_freq_hz]\" and enable_mapping = false." +
+		" Localization in offline mode is not supported.")
 )
 
 // Validate creates the list of implicit dependencies.
--- a/sensorprocess/movementsensorprocess.go
+++ b/sensorprocess/movementsensorprocess.go
@@ -13,85 +13,131 @@ import (
 	s "github.com/viamrobotics/viam-cartographer/sensors"
 )
 
-// StartIMU polls the IMU to get the next sensor reading and adds it to the cartofacade.
-// Stops when the context is Done.
-func (config *Config) StartIMU(ctx context.Context) {
+// StartMovementSensor polls the movement sensor to get the next sensor reading
+// and adds it to the cartofacade. Stops when the context is Done.
+func (config *Config) StartMovementSensor(ctx context.Context) {
 	for {
 		select {
 		case <-ctx.Done():
 			return
 		default:
-			if err := config.addIMUReadingInOnline(ctx); err != nil {
+			if err := config.addMovementSensorReadingInOnline(ctx); err != nil {
 				config.Logger.Warn(err)
 			}
 		}
 	}
 }
 
-// addIMUReadingInOnline ensures the most recent IMU scan, after corresponding lidar scans, gets processed by cartographer.
-func (config *Config) addIMUReadingInOnline(ctx context.Context) error {
-	// get next IMU data response; ignoring status since it is always false
-	imuReading, err := config.IMU.TimedMovementSensorReading(ctx)
+// addMovementSensorReadingInOnline ensures the most recent movement sensor scan,
+// after corresponding lidar scans, gets processed by cartographer.
+func (config *Config) addMovementSensorReadingInOnline(ctx context.Context) error {
+	// get next movement sensor data response; ignoring status since it is always false
+	movementSensorReading, err := config.MovementSensor.TimedMovementSensorReading(ctx)
 	if err != nil {
-		config.Logger.Warn(err)
 		if errors.Is(err, replaymovementsensor.ErrEndOfDataset) {
 			time.Sleep(1 * time.Second)
 		}
 		return err
 	}
 
-	// add IMU data to cartographer and sleep remainder of time interval
-	timeToSleep := config.tryAddIMUReadingOnce(ctx, *imuReading.TimedIMUResponse)
+	// add movement sensor data to cartographer and sleep remainder of time interval
+	timeToSleep := config.tryAddMovementSensorReadingOnce(ctx, movementSensorReading)
 
-	if !imuReading.TestIsReplaySensor {
+	if !movementSensorReading.TestIsReplaySensor {
 		time.Sleep(time.Duration(timeToSleep) * time.Millisecond)
-		config.Logger.Debugf("imu sleep for %vms", timeToSleep)
+		config.Logger.Debugf("movement sensor sleep for %vms", timeToSleep)
 	}
 
 	return nil
 }
 
-// tryAddIMUReadingUntilSuccess adds a reading to the cartofacade and retries on error (offline mode).
+// tryAddMovementSensorReadingUntilSuccess adds a reading to the cartofacade and retries on error (offline mode).
 // While add sensor reading fails, keep trying to add the same reading - in offline mode we want to
 // process each reading so if we cannot acquire the lock we should try again.
-func (config *Config) tryAddIMUReadingUntilSuccess(ctx context.Context, reading s.TimedIMUReadingResponse) error {
+func (config *Config) tryAddMovementSensorReadingUntilSuccess(ctx context.Context, reading s.TimedMovementSensorReadingResponse) error {
+	var imuDone, odometerDone bool
+	if !config.MovementSensor.Properties().IMUSupported {
+		imuDone = true
+	}
+	if !config.MovementSensor.Properties().OdometerSupported {
+		odometerDone = true
+	}
 	for {
 		select {
 		case <-ctx.Done():
 			return ctx.Err()
 		default:
-			if err := config.tryAddIMUReading(ctx, reading); err != nil {
-				if !errors.Is(err, cartofacade.ErrUnableToAcquireLock) {
-					config.Logger.Warnw("Retrying sensor reading due to error from cartofacade", "error", err)
+			if !imuDone {
+				if err := config.tryAddIMUReading(ctx, *reading.TimedIMUResponse); err != nil {
+					if !errors.Is(err, cartofacade.ErrUnableToAcquireLock) {
+						config.Logger.Warnw("Retrying IMU sensor reading due to error from cartofacade", "error", err)
+					}
+				} else {
+					imuDone = true
 				}
-			} else {
+			}
+			if !odometerDone {
+				if err := config.tryAddOdometerReading(ctx, *reading.TimedOdometerResponse); err != nil {
+					if !errors.Is(err, cartofacade.ErrUnableToAcquireLock) {
+						config.Logger.Warnw("Retrying odometer sensor reading due to error from cartofacade", "error", err)
+					}
+				} else {
+					odometerDone = true
+				}
+			}
+			if imuDone && odometerDone {
 				return nil
 			}
 		}
 	}
 }
 
-// tryAddIMUReadingOnce adds a reading to the carto facade and does not retry. Returns remainder of time interval.
-func (config *Config) tryAddIMUReadingOnce(ctx context.Context, reading s.TimedIMUReadingResponse) int {
+// tryAddMovementSensorReadingOnce adds a reading to the carto facade and does not retry. Returns remainder of time interval.
+func (config *Config) tryAddMovementSensorReadingOnce(ctx context.Context, reading s.TimedMovementSensorReadingResponse) int {
 	startTime := time.Now().UTC()
-	if err := config.tryAddIMUReading(ctx, reading); err != nil {
-		if errors.Is(err, cartofacade.ErrUnableToAcquireLock) {
-			config.Logger.Debugw("Skipping sensor reading due to lock contention in cartofacade", "error", err)
-		} else {
-			config.Logger.Warnw("Skipping sensor reading due to error from cartofacade", "error", err)
+
+	if config.MovementSensor.Properties().IMUSupported {
+		if err := config.tryAddIMUReading(ctx, *reading.TimedIMUResponse); err != nil {
+			if errors.Is(err, cartofacade.ErrUnableToAcquireLock) {
+				config.Logger.Debugw("Skipping IMU sensor reading due to lock contention in cartofacade", "error", err)
+			} else {
+				config.Logger.Warnw("Skipping IMU sensor reading due to error from cartofacade", "error", err)
+			}
 		}
 	}
+
+	if config.MovementSensor.Properties().OdometerSupported {
+		if err := config.tryAddOdometerReading(ctx, *reading.TimedOdometerResponse); err != nil {
+			if errors.Is(err, cartofacade.ErrUnableToAcquireLock) {
+				config.Logger.Debugw("Skipping odometer sensor reading due to lock contention in cartofacade", "error", err)
+			} else {
+				config.Logger.Warnw("Skipping odometer sensor reading due to error from cartofacade", "error", err)
+			}
+		}
+	}
+
 	timeElapsedMs := int(time.Since(startTime).Milliseconds())
-	return int(math.Max(0, float64(1000/config.IMU.DataFrequencyHz()-timeElapsedMs)))
+	return int(math.Max(0, float64(1000/config.MovementSensor.DataFrequencyHz()-timeElapsedMs)))
 }
 
-// tryAddIMUReading tries to add a reading to the carto facade.
+// tryAddIMUReading tries to add an IMU reading to the carto facade.
 func (config *Config) tryAddIMUReading(ctx context.Context, reading s.TimedIMUReadingResponse) error {
-	err := config.CartoFacade.AddIMUReading(ctx, config.Timeout, config.IMU.Name(), reading)
+	err := config.CartoFacade.AddIMUReading(ctx, config.Timeout, config.MovementSensor.Name(), reading)
 	if err != nil {
 		config.Logger.Debugf("%v \t |  IMU  | Failure \t \t | %v \n", reading.ReadingTime, reading.ReadingTime.Unix())
 	} else {
 		config.Logger.Debugf("%v \t |  IMU  | Success \t \t | %v \n", reading.ReadingTime, reading.ReadingTime.Unix())
+	}
+	return err
+}
+
+// tryAddOdometerReading tries to add an odometer reading to the carto facade.
+func (config *Config) tryAddOdometerReading(ctx context.Context, reading s.TimedOdometerReadingResponse) error {
+	err := config.CartoFacade.AddOdometerReading(ctx, config.Timeout, config.MovementSensor.Name(), reading)
+	if err != nil {
+		config.Logger.Debugf("%v \t |  Odometer  | Failure \t \t | %v \n", reading.ReadingTime, reading.ReadingTime.Unix())
+	} else {
+		config.Logger.Debugf("%v \t |  Odometer  | Success \t \t | %v \n", reading.ReadingTime, reading.ReadingTime.Unix())
 	}
 	return err
 }
--- a/sensorprocess/movementsensorprocess.go
+++ b/sensorprocess/movementsensorprocess.go
@@ -56,9 +56,11 @@ func (config *Config) addMovementSensorReadingInOnline(ctx context.Context) erro
 // process each reading so if we cannot acquire the lock we should try again.
 func (config *Config) tryAddMovementSensorReadingUntilSuccess(ctx context.Context, reading s.TimedMovementSensorReadingResponse) error {
 	var imuDone, odometerDone bool
+	// set IMU as done since it is not supported: we won't attempt to add IMU data to cartographer
 	if !config.MovementSensor.Properties().IMUSupported {
 		imuDone = true
 	}
+	// set odometer as done since it is not supported: we won't attempt to add odometer data to cartographer
 	if !config.MovementSensor.Properties().OdometerSupported {
 		odometerDone = true
 	}
--- a/sensorprocess/movementsensorprocess.go
+++ b/sensorprocess/movementsensorprocess.go
@@ -28,8 +28,8 @@ func (config *Config) StartMovementSensor(ctx context.Context) {
 	}
 }
 
-// addMovementSensorReadingInOnline ensures the most recent movement sensor scan,
-// after corresponding lidar scans, gets processed by cartographer.
+// addMovementSensorReadingInOnline attempts to get and add a movement sensor reading to the
+// cartofacade.
 func (config *Config) addMovementSensorReadingInOnline(ctx context.Context) error {
 	// get next movement sensor data response; ignoring status since it is always false
 	movementSensorReading, err := config.MovementSensor.TimedMovementSensorReading(ctx)
--- a/sensorprocess/movementsensorprocess.go
+++ b/sensorprocess/movementsensorprocess.go
@@ -67,15 +67,6 @@ func (config *Config) tryAddMovementSensorReadingUntilSuccess(ctx context.Contex
 		case <-ctx.Done():
 			return ctx.Err()
 		default:
-			if !imuDone {
-				if err := config.tryAddIMUReading(ctx, *reading.TimedIMUResponse); err != nil {
-					if !errors.Is(err, cartofacade.ErrUnableToAcquireLock) {
-						config.Logger.Warnw("Retrying IMU sensor reading due to error from cartofacade", "error", err)
-					}
-				} else {
-					imuDone = true
-				}
-			}
 			if !odometerDone {
 				if err := config.tryAddOdometerReading(ctx, *reading.TimedOdometerResponse); err != nil {
 					if !errors.Is(err, cartofacade.ErrUnableToAcquireLock) {
@@ -83,6 +74,15 @@ func (config *Config) tryAddMovementSensorReadingUntilSuccess(ctx context.Contex
 					}
 				} else {
 					odometerDone = true
+				}
+			}
+			if !imuDone {
+				if err := config.tryAddIMUReading(ctx, *reading.TimedIMUResponse); err != nil {
+					if !errors.Is(err, cartofacade.ErrUnableToAcquireLock) {
+						config.Logger.Warnw("Retrying IMU sensor reading due to error from cartofacade", "error", err)
+					}
+				} else {
+					imuDone = true
 				}
 			}
 			if imuDone && odometerDone {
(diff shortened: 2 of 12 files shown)


sensors/movementsensor.go
```diff
@@ -107,36 +108,36 @@ func (ms *MovementSensor) TimedMovementSensorReading(ctx context.Context) (Timed
 
 	timeoutCtx, cancel := context.WithTimeout(ctx, timedMovementSensorReadingTimeout)
 	defer cancel()
-	if ms.imuSupported {
-	imuLoop:
+	if ms.odometerSupported {
```
kkufieta, line 110: Adding odometer first, since we want to prioritize it over the imu

sensors/movementsensor.go
```diff
@@ -24,7 +24,7 @@ const (
 )
 
 var (
-	defaultTime = time.Time{}
+	undefinedTime = time.Time{}
```
kkufieta, line 27: Renamed this since I kept getting confused with `defaultTime`, assuming it was set to a specific time, when really it's just used to denote that it was not set to any meaningful value yet.

sensors/test_deps.go
```diff
@@ -415,3 +507,69 @@ func getMovementSensorWithInvalidProperties() *inject.MovementSensor {
...
+	movementSensor.LinearAccelerationFunc = func(ctx context.Context, extra map[string]interface{}) (r3.Vector, error) {
+		return r3.Vector{}, replay.ErrEndOfDataset
+	}
+	movementSensor.AngularVelocityFunc = func(ctx context.Context, extra map[string]interface{}) (spatialmath.AngularVelocity, error) {
+		return spatialmath.AngularVelocity{}, replay.ErrEndOfDataset
+	}
+	movementSensor.PositionFunc = func(ctx context.Context, extra map[string]interface{}) (*geo.Point, float64, error) {
+		return geo.NewPoint(0, 0), 0.0, replay.ErrEndOfDataset
+	}
+	movementSensor.OrientationFunc = func(ctx context.Context, extra map[string]interface{}) (spatialmath.Orientation, error) {
```
jeremyrhyde, line 563: nil

sensors/movementsensor.go
```diff
@@ -58,7 +58,7 @@ type TimedMovementSensorReadingResponse struct {
 
 // TimedIMUReadingResponse represents an IMU sensor reading with a time.
 type TimedIMUReadingResponse struct {
-	AngularVelocity    spatialmath.AngularVelocity
+	AngularVelocity    spatialmath.AngularVelocity // We set the values in radians/s instead of deg/s
```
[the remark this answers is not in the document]
jeremyrhyde: Can we add the same to odometer orientation?

sensorprocess/movementsensorprocess.go
```diff
@@ -13,85 +13,131 @@ import (
...
+			if !imuDone {
+				if err := config.tryAddIMUReading(ctx, *reading.TimedIMUResponse); err != nil {
+					if !errors.Is(err, cartofacade.ErrUnableToAcquireLock) {
+						config.Logger.Warnw("Retrying IMU sensor reading due to error from cartofacade", "error", err)
+					}
+				} else {
+					imuDone = true
+				}
+			}
+			if imuDone && odometerDone {
```
jeremyrhyde, line 88: [q] Do we want to stop only if both are done, or if either is done?

sensorprocess/sensorprocess.go
```diff
@@ -19,16 +20,48 @@ type Config struct {
...
+		!config.MovementSensor.Properties().OdometerSupported) {
+		return s.TimedMovementSensorReadingResponse{}, errors.New("movement sensor is not supported")
+	}
+	for {
+		movementSensorReading, err := config.MovementSensor.TimedMovementSensorReading(ctx)
+		if err != nil {
+			return s.TimedMovementSensorReadingResponse{}, err
+		}
+
+		var readingTime time.Time
```
jeremyrhyde, line 46: [opt] We could define the readingTime without the if statements via soemthing like:

`readingTime := movementSensorReading.TimedOdometerResponse.ReadingTime.Add(movementSensorReading.TimedOdometerResponse.ReadingTime.Sub(movementSensorReading.TimedIMUResponse.ReadingTime)) `

since if only one sensor is defined, it will be time.Time{} and if both, this function should give us the max time.

Not as clear as you have it so up to you if you want to try for a single definition

jeremyrhyde, line 46: [opt] We could define the readingTime without the if statements via soemthing like:

`readingTime := movementSensorReading.TimedOdometerResponse.ReadingTime.Add(movementSensorReading.TimedOdometerResponse.ReadingTime.Sub(movementSensorReading.TimedIMUResponse.ReadingTime)) `

since if only one sensor is defined, it will be time.Time{} and if both, this function should give us the max time.

Not as clear as you have it so up to you if you want to try for a single definition

sensorprocess/sensorprocess.go
```diff
@@ -37,32 +70,64 @@ func (config *Config) StartOfflineSensorProcess(ctx context.Context) bool {
...
-				break
-			}
+		movementSensorReading, err = config.getInitialMovementSensorReading(ctx, lidarReading)
+		if err != nil {
+			config.Logger.Warn(err)
+			return errors.Is(err, replaymovementsensor.ErrEndOfDataset)
 		}
 	}
 
+	type readingTime struct {
```
jeremyrhyde, line 84: [opt] define outside func at top of the file as offlineSensorReadingTime

sensorprocess/sensorprocess.go
```diff
@@ -37,32 +70,64 @@ func (config *Config) StartOfflineSensorProcess(ctx context.Context) bool {
...
+	}
+
 	// loop over all the data until one of the datasets has reached its end
 	for {
 		select {
 		case <-ctx.Done():
 			return false
 		default:
+			// create a map of supported sensors and their reading time stamps
+			readingTimes := []readingTime{
```
jeremyrhyde, line 96: could also move the struct definition to here so its 

```
readingTimes = []struct{
         sensorType  string
         readingTime time.Time
} {
       {sensorType: "lidar", readingTime: lidarReading.ReadingTime},
}
```

jeremyrhyde, line 96: could also move the struct definition to here so its 

```
readingTimes = []struct{
         sensorType  string
         readingTime time.Time
} {
       {sensorType: "lidar", readingTime: lidarReading.ReadingTime},
}
```

sensorprocess/sensorprocess.go
jeremyrhyde: see below for another alternative

sensorprocess/sensorprocess.go
```diff
@@ -37,32 +70,64 @@ func (config *Config) StartOfflineSensorProcess(ctx context.Context) bool {
...
+				{sensorType: "lidar", readingTime: lidarReading.ReadingTime},
+			}
+			// default to the slightly later imu timestamp, in case that the odometer time stamp was
+			// taken before the lidar time stamp, but the imu time stamp was taken after the lidar time
+			// stamp, we'll want to prioritize adding the lidar measurement before adding the movement
+			// sensor measurement
+			if config.MovementSensor != nil && config.MovementSensor.Properties().IMUSupported {
```
jeremyrhyde, line 103: [opt] Could always add odom and imu reading times even if they are not supported by properties and then have your sort function put timestamps of time.Time{} at end

sensorprocess/sensorprocess.go
```diff
@@ -37,32 +70,64 @@ func (config *Config) StartOfflineSensorProcess(ctx context.Context) bool {
...
+				readingTimes = append(readingTimes,
+					readingTime{
+						sensorType:  "movement-sensor",
+						readingTime: movementSensorReading.TimedOdometerResponse.ReadingTime,
+					})
+			}
+
+			// sort the readings based on their time stamp
+			sort.Slice(readingTimes,
+				func(i, j int) bool {
```
jeremyrhyde, line 119: [opt/nit] put in separate function

sensorprocess/sensorprocess.go
```diff
@@ -37,32 +70,64 @@ func (config *Config) StartOfflineSensorProcess(ctx context.Context) bool {
...
+			// default to the slightly later imu timestamp, in case that the odometer time stamp was
+			// taken before the lidar time stamp, but the imu time stamp was taken after the lidar time
+			// stamp, we'll want to prioritize adding the lidar measurement before adding the movement
+			// sensor measurement
+			if config.MovementSensor != nil && config.MovementSensor.Properties().IMUSupported {
+				readingTimes = append(readingTimes,
+					readingTime{
+						sensorType:  "movement-sensor",
+						readingTime: movementSensorReading.TimedIMUResponse.ReadingTime,
+					})
```
jeremyrhyde, line 108: [q] if both imu and odom are available, will they have the same timestamp?

sensorprocess/sensorprocess_test.go
```diff
@@ -28,12 +29,21 @@ func TestStartOfflineSensorProcess(t *testing.T) {
 		ReadingTime: time.Now().UTC(),
 	}
 
+	odometerReading := s.TimedOdometerReadingResponse{
+		Position:    geo.NewPoint(5, 4),
```
jeremyrhyde, line 33: [q] arent these values defined in another file? If so they should be referenced from there

sensorprocess/sensorprocess_test.go
```diff
@@ -106,48 +137,89 @@ func TestStartOfflineSensorProcess(t *testing.T) {
...
 			imuEnabled              bool
+			odometerEnabled         bool
 			lidarReadingTimeAddedMs []int
 			msReadingTimeAddedMs    []int
 			expectedDataInsertions  []string
 		}{
 			{
-				description:             "no imu data is added if imu is not enabled",
+				description:             "no movement sensor data is added if movement sensor is not enabled",
+				odometerEnabled:         false,
```
jeremyrhyde, line 147: [nit] move under imuEnabled

sensorprocess/sensorprocess_test.go
```diff
@@ -160,14 +232,30 @@ func TestStartOfflineSensorProcess(t *testing.T) {
...
+						if tt.imuEnabled {
+							movementSensorReading.TimedIMUResponse = &imuReading
+							imuReadingTime := now.Add(time.Duration(tt.msReadingTimeAddedMs[numMovementSensorData]) * time.Millisecond)
+							movementSensorReading.TimedIMUResponse.ReadingTime = imuReadingTime
+						}
+						numMovementSensorData++
 						return movementSensorReading, nil
 					}
-					return s.TimedMovementSensorReadingResponse{}, replay.ErrEndOfDataset
+					return movementSensorReading, replay.ErrEndOfDataset
```
jeremyrhyde, line 252: [q] why are no longer returning an empty s.TimedMovementSensorReadingResponse{}?

sensorprocess/sensorprocess_test.go
```diff
@@ -202,506 +299,171 @@ func TestStartOfflineSensorProcess(t *testing.T) {
...
-	injectLidar.DataFrequencyHzFunc = func() int { return dataFrequencyHz }
-
-	config := Config{
-		Logger:      logger,
-		CartoFacade: &cf,
-		IsOnline:    injectLidar.DataFrequencyHzFunc() != 0,
-		Lidar:       &injectLidar,
-		Timeout:     10 * time.Second,
+	odometerReading := s.TimedOdometerReadingResponse{
+		Position:    geo.NewPoint(5, 4),
```
jeremyrhyde, line 324: Values defined in other file

sensorprocess/sensorprocess_test.go
jeremyrhyde: Note: If youd prefer to have a separate def, id define these at the top of the file as they are used in multiple fucntions

sensorprocess/testhelper.go
```diff
@@ -120,19 +123,55 @@ func onlineModeLidarTestHelper(
 	}
 }
 
-func onlineModeIMUTestHelper(
+func invalidAddLidarReadingInOnlineTestHelper(
```
jeremyrhyde, line 126: [q] what makes this function invalid? The addLidarReadingFunc returns a nil err every time

sensorprocess/testhelper.go
```diff
@@ -144,104 +183,172 @@ func onlineModeIMUTestHelper(
...
-		// the IMU test fixture happens to always return the same readings currently
-		// in reality they are likely different every time
-		test.That(t, call.currentReading.LinearAcceleration, test.ShouldResemble, expectedIMUReading.LinearAcceleration)
-		test.That(t, call.currentReading.AngularVelocity, test.ShouldResemble, expectedIMUReading.AngularVelocity)
-		test.That(t, call.timeout, test.ShouldEqual, config.Timeout)
+	if movementSensor.Properties().IMUSupported {
+		for i, call := range imuCalls {
+			t.Logf("call %d", i)
+			test.That(t, call.sensorName, test.ShouldResemble, string(testMovementSensor))
+			// the IMU test fixture happens to always return the same readings currently
```
jeremyrhyde, line 251:
```suggestion
			// the IMU test fixture happens to always return the same readings currently.
```

[nit] minor but i read it first as ... readings currently in reality which confused me

sensorprocess/testhelper.go
```diff
@@ -144,104 +183,172 @@ func onlineModeIMUTestHelper(
...
-		readingTime, err := time.Parse(time.RFC3339Nano, s.TestTimestamp)
-		test.That(t, err, test.ShouldBeNil)
-		test.That(t, calls[0].currentReading.ReadingTime.Equal(readingTime), test.ShouldBeTrue)
-	} else {
-		t.Errorf("no timestamp tests provided for %v", string(testImu))
+	if movementSensor.Properties().OdometerSupported {
+		for i, call := range odometerCalls {
+			t.Logf("call %d", i)
+			test.That(t, call.sensorName, test.ShouldResemble, string(testMovementSensor))
+			// the odometer test fixture happens to always return the same readings currently
```
jeremyrhyde, line 267: [nit] same

sensorprocess/movementsensorprocess.go
```diff
@@ -13,85 +13,131 @@ import (
 	s "github.com/viamrobotics/viam-cartographer/sensors"
```
jeremyrhyde, line 13: Very thorough testing!!!

sensors/test_deps.go
```diff
@@ -356,6 +381,48 @@ func getOdometerWithErroringFunctions() *inject.MovementSensor {
...
+			OrientationSupported: true,
+		}, nil
+	}
+	return odometer
+}
+
+func getFinishedReplayOdometer() *inject.MovementSensor {
+	odometer := &inject.MovementSensor{}
+	odometer.PositionFunc = func(ctx context.Context, extra map[string]interface{}) (*geo.Point, float64, error) {
+		return geo.NewPoint(0, 0), 0.0, replay.ErrEndOfDataset
```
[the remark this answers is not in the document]
randhid: I prefer `math.Nan` for things that actually mean something physical.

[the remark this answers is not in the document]
jeremyrhyde: So you would want `geo.NewPoint(math.Nan, math.Nan)`? I suggested nil because the return of Position is a `geo.Point` pointer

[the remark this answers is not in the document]
jeremyrhyde: Note: I also just checked RDK and movement sensors that dont implement Position (such as imuwit) returns geo.NewPoint(0,0). This isn't my favorite pattern to support as if an error is returned I'd like the result to be inaccessible and not take up any memory (so nil)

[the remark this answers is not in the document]
randhid: ^ No, not all; our preferred pattern with an error is `math.Nan`. We just haven't gotten around to changing them yet. So you're looking at a bad example. Also the ones that truly don't have out ErrUnimplemented returned, so you can't actually get info out of that client call at all.

The one that does implement Position and has an error associated with it is GPS, where if it is erroring, we return math.NaN. 

Also the problem here is if you return it as nil, and someone else uses the geoPoint library that we're wrapping on that nil pointer - like `position.MidPointTo` you get panics all over the place. I'd rather give NaNs.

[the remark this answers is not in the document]
randhid: The different here is - the witimu can never return position - it doesn't have that data or a chance at returning that data at all, so we use the movement sensor error unimplemented and a PositionReporting: false Property to make sure that doesn't show up in the sdks or the rc card. So you just get an error unimplemented if you try to get any of that information for business level logic out of the wit imu.

sensorprocess/movementsensorprocess.go
randhid: I think this is covered by lines 58-64, where she's setting them to done if the properties are not supported, but maybe a comment to explain this kat.

randhid (changes requested):
Some comments would be good in the places I note, also I'd like the case selector to be a tad bit neater, and why do we need that `closed` bool?

Edit: I'm not looking at the tests at all, but I trust they're doing good things! Lots of them!

sensorprocess/movementsensorprocess.go
```diff
@@ -13,85 +13,131 @@ import (
...
 	timeElapsedMs := int(time.Since(startTime).Milliseconds())
-	return int(math.Max(0, float64(1000/config.IMU.DataFrequencyHz()-timeElapsedMs)))
+	return int(math.Max(0, float64(1000/config.MovementSensor.DataFrequencyHz()-timeElapsedMs)))
 }
 
-// tryAddIMUReading tries to add a reading to the carto facade.
+// tryAddIMUReading tries to add an IMU reading to the carto facade.
 func (config *Config) tryAddIMUReading(ctx context.Context, reading s.TimedIMUReadingResponse) error {
-	err := config.CartoFacade.AddIMUReading(ctx, config.Timeout, config.IMU.Name(), reading)
+	err := config.CartoFacade.AddIMUReading(ctx, config.Timeout, config.MovementSensor.Name(), reading)
```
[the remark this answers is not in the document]
kkufieta: Tested it, it's currently just the short name:

```bash
config.MovementSensor.Name():  replay-odom
```

What is preferred, and what are the consequences of using one vs. the other?

viam_cartographer.go
```diff
@@ -395,25 +393,30 @@ func initCartoFacade(ctx context.Context, cartoSvc *CartographerService) error {
...
-		EnableMapping:      cartoSvc.enableMapping,
-		ExistingMap:        cartoSvc.existingMap,
-	}
-
-	if cartoSvc.movementSensorName == "" {
+	var movementSensorName string
+	if cartoSvc.movementSensor == nil {
 		cartoSvc.logger.Debug("No movement sensor provided, setting use_imu_data to false")
 	} else {
+		movementSensorName = cartoSvc.movementSensor.Name()
```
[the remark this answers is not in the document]
kkufieta: We're calling here `Name()` as defined in [viam-cartographer/sensors/movementsensor.go](https://github.com/viamrobotics/viam-cartographer/blob/main/sensors/movementsensor.go#L84). That name is set from the optional parameters in [viam-cartographer/config/config.go](https://github.com/viamrobotics/viam-cartographer/blob/main/config/config.go#L96), which is fed into slam via `"config_params"`, e.g. like this:

```json
  "services": [
    {
      "name": "myslam",
      "type": "slam",
      "attributes": {
        "enable_mapping": true,
        "use_cloud_slam": false,
        "camera": {
          "name": "replay-lidar",
          "data_frequency_hz": "5"
        },
        "config_params": {
          "mode": "2d",
          "max_range_meters": "12",
          "min_range_meters": "0.3"
        },
        "movement_sensor": {
          "name": "replay-odom",
          "data_frequency_hz": "20"
        }
      },
      "model": "viam:slam:cartographer"
    }
  ]
```

sensorprocess/movementsensorprocess.go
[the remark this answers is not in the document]
randhid: Short name is preferred as it usually matches the name of the dependency that you're passing into the config. The full name is prepended with more information abut the namespace and family, which could be anything for a module.

I was just thrown off because the resource manager's API uses `Name` for the full name and `ShortName` of the shortname, so needed to make sure what this was.

viam_cartographer.go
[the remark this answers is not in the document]
randhid: Cool, thanks for that explanation.

sensors/test_deps.go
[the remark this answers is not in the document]
randhid: ^ Since those aren't returned pointers like the *geo.Point return from Position: I'm not as concerned about the r3 library accessing them after they're returned. So it is up to you in this pr.

 Is cartographer seeing zero values and taking them as meaningful? That's a question for another day, maybe it has a safeguard against a constant stream of zero data, some algorithms do.

viam_cartographer.go
```diff
@@ -187,32 +187,31 @@ func New(
 		return nil, err
 	}
 
+	// Get the lidar for the Dim2D cartographer sub algorithm
 	lidarName := svcConfig.Camera["name"]
```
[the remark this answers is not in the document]
randhid: I'm realizing I'm used to the styling in rdk  - if the config has a Camera, I assume it to be something like:
```go
Camera {
     Name string. `json:"name"`
     DataFrequency int `json:"data_frequency_hz,omitempty"`
}
```

So it's cleaner to validate and access and you don't have to remember the key in the string all over the place

but now I see in `config.go` the attributes are maps rather than struts that can be serialized. I don't like it but it's a minor cleanup point. Don't do in this pr.

Okay to leave as is.

sensorprocess/sensorprocess.go
kkufieta: Don't understand; let's discuss this in person

viam_cartographer.go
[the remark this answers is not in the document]
kkufieta: Should we create a ticket for this?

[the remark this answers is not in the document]
randhid: No, it's different, but I don't really care that it's different. Even if I like the in rdk styling more.

viam_cartographer.go
```diff
@@ -500,7 +503,11 @@ func (cartoSvc *CartographerService) Position(ctx context.Context) (spatialmath.
 		cartoSvc.logger.Warn("Position called with use_cloud_slam set to true")
 		return nil, "", ErrUseCloudSlamEnabled
 	}
-	if cartoSvc.closed {
+
+	cartoSvc.muClose.Lock()
+	cartoSvcClosed := cartoSvc.closed
```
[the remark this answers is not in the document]
jeremyrhyde: Then is your preference to save cartoSvc.closed as var cartoSvcClosed purely to reduce the number of Unlock statements? I personally would prefer:

```
cartoSvc.muClose.Lock()
	if cartoSvc.closed {
		cartoSvc.logger.Warn("Position called after shutting down of cartographer has been initiated")
                 cartoSvc.muClose.Unlock()
		return nil, "", ErrClosed
	}
cartoSvc.muClose.Unlock()
```

as the new variable definition prompted me to look into if it was being used elsewhere in the function. But I dont feel too strongly.

sensors/test_deps.go
```diff
@@ -81,10 +81,20 @@ const (
 	// OdometerWithErroringFunctions is an Odometer whose functions return errors.
 	OdometerWithErroringFunctions TestSensor = "odometer_with_erroring_functions"
 
+	// ReplayOdometer is an odometer that works as expected and returns position and orientation values.
```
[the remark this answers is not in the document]
jeremyrhyde: Thank you!

sensors/test_deps.go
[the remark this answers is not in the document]
jeremyrhyde: TBR (To be researched) about carto's handling of constant streams of zeros

sensors/test_deps.go
```diff
@@ -389,6 +456,31 @@ func getMovementSensorBothIMUAndOdometer() *inject.MovementSensor {
...
+func getMovementSensorBothIMUAndOdometerWithErroringFunctions() *inject.MovementSensor {
+	movementSensor := &inject.MovementSensor{}
+	movementSensor.LinearAccelerationFunc = func(ctx context.Context, extra map[string]interface{}) (r3.Vector, error) {
+		return r3.Vector{}, errors.New(InvalidSensorTestErrMsg)
+	}
+	movementSensor.AngularVelocityFunc = func(ctx context.Context, extra map[string]interface{}) (spatialmath.AngularVelocity, error) {
+		return spatialmath.AngularVelocity{}, errors.New(InvalidSensorTestErrMsg)
+	}
+	movementSensor.PositionFunc = func(ctx context.Context, extra map[string]interface{}) (*geo.Point, float64, error) {
+		return &geo.Point{}, 0.0, errors.New(InvalidSensorTestErrMsg)
```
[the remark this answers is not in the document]
jeremyrhyde: Based on that comment using math.Nan seems correct! (note your comment here has the wrong err, but the code is right)

sensors/movementsensor.go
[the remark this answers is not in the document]
jeremyrhyde: Gotcha, because orientation is expressed as a quaternion correct?

sensorprocess/movementsensorprocess.go
jeremyrhyde: What about the instance of an offline run where, during data generation, the odom got disconnected but the movement sensor kept gathering data for some time? Waiting for both may cause cartographer to attempt to add 1000s of IMU data points after odom data is complete.

Another example of this could be if a user is lax in their time interval definition and accidentally grabs data from a different run which only had an odom movement sensor

sensorprocess/sensorprocess.go
jeremyrhyde: Right, that was the whole GetReadings fiasco

sensorprocess/testhelper.go
jeremyrhyde: I'd prefer to remove the invalid if a good lidar COULD be passed into the function and result in valid information. That reduces the confusion for:

- `invalidAddLidarReadingInOnlineTestHelper(invalidLiDAR) -> invalid result`
- `invalidAddLidarReadingInOnlineTestHelper(validLiDAR) -> valid result <-------`

to

- `addLidarReadingInOnlineTestHelper(invalidLiDAR) -> invalid result`
- `addLidarReadingInOnlineTestHelper(validLiDAR) -> valid result`

viam_cartographer.go
```diff
@@ -463,13 +466,13 @@ func terminateCartoFacade(ctx context.Context, cartoSvc *CartographerService) er
...
-	movementSensor     s.TimedMovementSensor
-	movementSensorName string
-	subAlgo            SubAlgo
+	SlamMode       cartofacade.SlamMode
+	lidar          s.TimedLidar
+	movementSensor s.TimedMovementSensor
+	subAlgo        SubAlgo
+
+	muClose sync.Mutex
+	closed  bool
```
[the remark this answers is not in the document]
jeremyrhyde: Note Rand, This came up in an eng talk by Tess six months back. Someone told me that the mod manager should be handling this now, but when I talked to Benji, he had no idea what I was talking about and didn't know if it was being handled so we decided to keep this closed check in, until we know for certain.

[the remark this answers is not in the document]
randhid: There was a slakc thread about this, it's not mod manager, it's the resource graph that handles removing the resource nodes from the graph (modules included). Its not something other modules do, since if it's closed the client only returns an error for a resource that's not there anymore.

Like if you try to import the resource from the robot after it's closed and not built again, the `FromRobot`, `FromDep` call in your code just fails. If it doesn't, that's a netcode bug.

viam_cartographer.go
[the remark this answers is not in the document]
jeremyrhyde: spoke offline

sensors/movementsensor.go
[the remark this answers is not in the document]
jeremyrhyde: spoke offline

sensorprocess/sensorprocess.go
jeremyrhyde: spoke ofline

sensorprocess/movementsensorprocess.go
jeremyrhyde: @randhid for context, we had this issue in the past where leftover imu data would cause the session to hang until timeout occurred.

After running through the full pipeline for offline mode, I believe this is handled by the new offline architecture, but wanted to make you aware of it in case it crops up again

config/config.go
```diff
@@ -38,7 +38,7 @@ type OptionalConfigParams struct {
 
 var (
 	errCameraMustHaveName        = errors.New("\"camera[name]\" is required")
-	errLocalizationInOfflineMode = newError("camera[data_freq_hz] and enable_mapping = false. localization in offline mode not supported.")
+	errLocalizationInOfflineMode = newError("\"camera[data_freq_hz]\" and enable_mapping = false. localization in offline mode not supported.")
```
jeremyrhyde, line 41:
```suggestion
	errLocalizationInOfflineMode = newError("\"camera[data_freq_hz]\" and enable_mapping = false. Localization in offline mode not supported.")
```

viam_cartographer.go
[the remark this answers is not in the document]
jeremyrhyde: Great! Then we don't need the `closed` or `closedMu`! @randhid should we handle that in this PR or create a ticket to be solved now (this PR is already pretty big, but based on that, the only change would need to be code removal)

jeremyrhyde (approved):
LGTM! see one or two last comments. @randhid can decide if we want to remove the `closed` checks in each endpoints in this PR or a future one.

viam_cartographer.go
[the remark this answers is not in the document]
randhid: So the only case to keep this is if cloud slam is directly importing cartographer methods outside of the resource graph.

For example, `gpsnmea` has public functions that are imported directly from within rdk by `gpstrk` (this is something we want to fix but never have bandwidth to), so we have a closed bool there to account for one structure being imported directly outside of the resource graph logic. 

If nothing is doing so with this module then it's unnecessary, but I want to be exhaustive about the potential gotchas.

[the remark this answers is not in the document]
randhid: Kat, have you seen any of these errors or logs surface at all? Is the C++ layer taking a long time to shut down and thus we are in danger of calling an endpoint directly before the resource graph removes this resource node?

sensorprocess/testhelper.go
kkufieta: It has to be "invalid" because it actually ensures that things fail. There's another, similar function, where we check that it's valid. Won't change anything now since refactoring this test is not part of the scope for this ticket, but will keep it in mind when I work on integration tests.

viam_cartographer.go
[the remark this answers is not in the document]
kkufieta: I have neither paid attention nor noticed it. Can we create a separate ticket to look into this?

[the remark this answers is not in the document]
randhid: Then let's not add code that might not do anything at all and wasn't hurting the implementation, but does seem to be important for anyone reading the code.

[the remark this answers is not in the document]
randhid: I was referring to your mutexes, okay to remove the closed checks to another pr.

sensorprocess/movementsensorprocess.go
kkufieta: I don't think this is going to be a problem. Since we support using both odom and imu together only via merged movement sensors, if one runs out of data, we get an`ErrEndOfDataset` error.  (see [`TimedMovementSensorReading`](https://github.com/kkufieta/viam-cartographer/blob/kk/rsdk-4862-sensorprocess-2/sensors/movementsensor.go#L95). If that happens, we [exit out](https://github.com/kkufieta/viam-cartographer/blob/kk/rsdk-4862-sensorprocess-2/sensorprocess/sensorprocess.go#L157) and don't call this function.

randhid (approved):

PR #308, opened 2023-12-13
[RSDK-5560] tweak replay sensor process

JohnN193:
https://viam.atlassian.net/browse/RSDK-5560

testing for adjusting the online replay sensor process to reduce the lag buildup in cloudslam. User's will see changes in their map within 5-10 seconds, and can be configured to be shorter by adjusting the sync rate in the user's capture config.

### DIFF
--- a/sensorprocess/lidarsensorprocess.go
+++ b/sensorprocess/lidarsensorprocess.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/viamrobotics/viam-cartographer/cartofacade"
 	s "github.com/viamrobotics/viam-cartographer/sensors"
+	"go.viam.com/rdk/components/camera/replaypcd"
 )
 
 // StartLidar polls the lidar to get the next sensor reading and adds it to the cartofacade.
@@ -32,6 +33,9 @@ func (config *Config) addLidarReadingInOnline(ctx context.Context) error {
 	// get next lidar data response; ignoring status since it is always false
 	lidarReading, err := config.Lidar.TimedLidarReading(ctx)
 	if err != nil {
+		if errors.Is(err, replaypcd.ErrEndOfDataset) {
+			time.Sleep(1 * time.Second)
+		}
 		return err
 	}
 
--- a/sensorprocess/lidarsensorprocess.go
+++ b/sensorprocess/lidarsensorprocess.go
@@ -37,8 +37,10 @@ func (config *Config) addLidarReadingInOnline(ctx context.Context) error {
 
 	// add lidar data to cartographer and sleep remainder of time interval
 	timeToSleep := config.tryAddLidarReadingOnce(ctx, lidarReading)
-	time.Sleep(time.Duration(timeToSleep) * time.Millisecond)
-	config.Logger.Debugf("lidar sleep for %vms", timeToSleep)
+	if !lidarReading.TestIsReplaySensor {
+		time.Sleep(time.Duration(timeToSleep) * time.Millisecond)
+		config.Logger.Debugf("lidar sleep for %vms", timeToSleep)
+	}
 	return nil
 }
 
--- a/sensorprocess/lidarsensorprocess.go
+++ b/sensorprocess/lidarsensorprocess.go
@@ -7,9 +7,10 @@ import (
 	"math"
 	"time"
 
+	"go.viam.com/rdk/components/camera/replaypcd"
+
 	"github.com/viamrobotics/viam-cartographer/cartofacade"
 	s "github.com/viamrobotics/viam-cartographer/sensors"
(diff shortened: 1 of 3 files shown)


sensorprocess/lidarsensorprocess.go
```diff
@@ -32,13 +33,18 @@ func (config *Config) addLidarReadingInOnline(ctx context.Context) error {
 	// get next lidar data response; ignoring status since it is always false
 	lidarReading, err := config.Lidar.TimedLidarReading(ctx)
 	if err != nil {
+		if errors.Is(err, replaypcd.ErrEndOfDataset) {
```
jeremyrhyde, line 36: [note to self] this sleep just to prevent overzealous calling of replay sensor when data will only be synced every sync interval

sensorprocess/lidarsensorprocess.go
```diff
@@ -32,13 +33,18 @@ func (config *Config) addLidarReadingInOnline(ctx context.Context) error {
...
+			time.Sleep(1 * time.Second)
+		}
 		return err
 	}
 
 	// add lidar data to cartographer and sleep remainder of time interval
 	timeToSleep := config.tryAddLidarReadingOnce(ctx, lidarReading)
-	time.Sleep(time.Duration(timeToSleep) * time.Millisecond)
-	config.Logger.Debugf("lidar sleep for %vms", timeToSleep)
+	if !lidarReading.TestIsReplaySensor {
```
jeremyrhyde, line 44: [q] I thought this sleep was being removed/ reduced to something like 10 ms?

jeremyrhyde (commented):
Looks good, but why is the sleep still there? Is this PR not intended to handle the replay lag?

sensorprocess/lidarsensorprocess.go
JohnN193: yeah, having the pause here helps reduce the amount of times we call `BinaryDataByFilter` to prevent us from overloading app

sensorprocess/lidarsensorprocess.go
jeremyrhyde: [note] used in localization mode and if a user forces use_cloud_slam to false during mapping. Was previous unused due to us treating online and offline identically

jeremyrhyde (approved):

sensorprocess/lidarsensorprocess.go
```diff
@@ -7,6 +7,8 @@ import (
 	"math"
 	"time"
 
+	"go.viam.com/rdk/components/camera/replaypcd"
```
randhid, line 10: from what I can tell, this introduces a dependency on rdk purely for that `ErrorEndOfDataset` export. Is there a way we can forego this completely? Can we just have identical errors instead of an exported variable? That's more maintenance work for now, but cleaner for the future.

randhid: I'd like use to always think about not cross-depending too much on rdk, helps far future us. 

I realize in some cases in this repo it's not possible - spatial math, RImage, etc. So would be okay to merge if you feel like this is overkill.

randhid (approved):

sensorprocess/lidarsensorprocess.go
JohnN193: since we own the replay sensors we could hardcode this error on the cartographer side. I think thats reasonable since we could already accidentally break this check if the version of viam-server used is different from the rdk version being imported here.

JohnN193: I think for now I will leave it as is, but maybe we can think more about how to handle rdk dependencies that might break us

sensorprocess/lidarsensorprocess.go
JohnN193: *unused due to us treating replay sensors the same as live sensors

sensorprocess/lidarsensorprocess.go
randhid: Cool! alright with me!

sensorprocess/lidarsensorprocess.go
kkufieta: @JohnN193 do we want this to use as well `strings.Contains(err.Error(), replaypcd.ErrEndOfDataset.Error())`?

JohnN193: Was this not already changed in https://github.com/viamrobotics/viam-cartographer/pull/312/files

kkufieta: I think only the cases in `sensorprocess.go` were changed - not the ones in `movementsensorprocess.go` and `lidarsensorprocess.go`

PR #309, opened 2023-12-18
Reduce Appimage Size

jeremyrhyde:
Tested arm64 appiamge on bookworm pi

### DIFF
--- a/.github/workflows/deploy_module.yml
+++ b/.github/workflows/deploy_module.yml
@@ -20,11 +20,11 @@ jobs:
       matrix:
         include:
           - arch: buildjet-8vcpu-ubuntu-2204
-            image: ghcr.io/viamrobotics/canon:amd64-cache
-            platform: amd64
+            image: ghcr.io/viamrobotics/rdk-devenv:amd64-cache
+            platform: linux/amd64
           - arch: buildjet-8vcpu-ubuntu-2204-arm
-            image: ghcr.io/viamrobotics/canon:arm64-cache
-            platform: arm64
+            image: ghcr.io/viamrobotics/rdk-devenv:arm64-cache
+            platform: linux/arm64
     runs-on: ${{ matrix.arch }}
     container:
       image: ${{ matrix.image }}
--- a/.github/workflows/deploy_module.yml
+++ b/.github/workflows/deploy_module.yml
@@ -21,10 +21,10 @@ jobs:
         include:
           - arch: buildjet-8vcpu-ubuntu-2204
             image: ghcr.io/viamrobotics/rdk-devenv:amd64-cache
-            platform: linux/amd64
+            platform: amd64
           - arch: buildjet-8vcpu-ubuntu-2204-arm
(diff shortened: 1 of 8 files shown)


etc/packaging/appimages/cartographer-module-aarch64.yml
```diff
@@ -26,58 +26,19 @@ AppDir:
     - arm64
     allow_unauthenticated: true
     sources:
-    - sourceline: deb [trusted=yes] http://deb.debian.org/debian bullseye main
-    - sourceline: deb [trusted=yes] http://deb.debian.org/debian-security bullseye-security main
-    - sourceline: deb [trusted=yes] http://deb.debian.org/debian bullseye-updates main
-    - sourceline: deb [trusted=yes] https://us-apt.pkg.dev/projects/static-file-server-310021 bullseye main
+    - sourceline: deb [trusted=yes] http://deb.debian.org/debian bookworm main
```
jeremyrhyde, line 29: [note] Upgrading to bookworm to use non-dev packages which are smaller and unavailable in bullseye. This upgrade also matches the new red-env used in workflows. Appimage will be backward compatible to bullseye

etc/packaging/appimages/cartographer-module-aarch64.yml
```diff
@@ -26,58 +26,19 @@ AppDir:
...
-    - libpcl-tracking1.11:arm64
-    - libpcl-visualization1.11:arm64
-    - libx11-6:arm64
-    - libflann1.9:arm64
-    - libxcb1:arm64
-    - libglx0:arm64
-    - libgl1:arm64
-    - libglvnd0:arm64
-    - libnlopt0:arm64
+    - libabsl20220623
```
abe-winter, line 34: out of scope for review, but you can probably use a single appimage yml for both platforms

they support environment variable in these, with `${VAR}` syntax I believe, and the diff between these files seems fairly small now

(totally optional, only if it matters to you)

.github/workflows/deploy_module.yml
```diff
@@ -20,11 +20,11 @@ jobs:
       matrix:
         include:
           - arch: buildjet-8vcpu-ubuntu-2204
-            image: ghcr.io/viamrobotics/canon:amd64-cache
-            platform: amd64
+            image: ghcr.io/viamrobotics/rdk-devenv:amd64-cache
+            platform: linux/amd64
```
abe-winter, line 24: this may fail when you run it -- this is being consumed elsewhere in the file as `matrix.platform`, both here + in the `upload` job

abe-winter (approved):
lgtm, see comments for single nit

etc/packaging/appimages/cartographer-module-aarch64.yml
jeremyrhyde: We'll handle this in the future, when a larger overhaul of the system occurs

.github/workflows/deploy_module.yml
jeremyrhyde: Made consistent with other workflows

PR #311, opened 2023-12-29
RSDK-6179 - use docommand to postprocess

kim-mishra:
This PR adds postprocessing functionality via the `DoCommand`. This is achieved by storing 2 additional parameters on the cartoSvc struct. 
1. postprocessed: which keeps track of if post processing is on or off
2. postprocessingTasks: a list of all of the postprocessing steps that have occurred. 

In `PointCloudMap` if postprocessing has been "turned on" (`cartoSvc.postprocessed == true`) all of the postprocessing tasks are iterated through & a new pointcloud map is returned with the added/removed points.

The following functionality was added to the `DoCommand`:
* `postprocess_toggle`: turns on and off postprocessing
* `postprocess_add`: adds points to pointcloud map. Value must be specified as `[{X: x, Y: y}, ...]`
* `postprocess_add`: adds points to pointcloud map. Value must be specified as `[{X: x, Y: y}, ...]`
* `postprocess_undo`: undoes last postprocessing steps.


cartofacade/capi.go
```diff
@@ -7,7 +7,7 @@ package cartofacade
 	#cgo CFLAGS: -I../viam-cartographer/src/carto_facade
 
 	// the libraries that need to be linked can be derived from line 258 of the build.ninja file that is autogenerated during make build
-	#cgo LDFLAGS: -lviam-cartographer  -lcartographer -ldl -lm -labsl_hash  -labsl_city -labsl_bad_optional_access -labsl_strerror  -labsl_str_format_internal -labsl_synchronization -labsl_strings -labsl_throw_delegate -lcairo -llua5.3 -lstdc++ -lceres -lprotobuf -lglog -lboost_filesystem -lboost_iostreams -lpcl_io -lpcl_common -labsl_raw_hash_set
+	#cgo LDFLAGS: -Wl,-rpath,/opt/homebrew/lib -lviam-cartographer  -lcartographer -ldl -lm -labsl_hash  -labsl_city -labsl_bad_optional_access -labsl_strerror  -labsl_str_format_internal -labsl_synchronization -labsl_strings -labsl_throw_delegate -lcairo -llua5.3 -lstdc++ -lceres -lprotobuf -lglog -lboost_filesystem -lboost_iostreams -lpcl_io -lpcl_common -labsl_raw_hash_set
```
kim-mishra, line 10: I had to add this to get it to build my local version

postprocess/postprocess.go
```diff
@@ -0,0 +1,161 @@
...
+		}
+
+		x := val.MapIndex(reflect.ValueOf("X"))
+		y := val.MapIndex(reflect.ValueOf("Y"))
+		task.Points = append(task.Points, r3.Vector{X: x.Elem().Float(), Y: y.Elem().Float()})
+	}
+
+	task.Instruction = instruction
+	return task, nil
+}
```
kim-mishra, line 63: This is all somewhat annoying/janky because the `DoCommand` expects an `interface{}` and I want a somewhat complex data structure as input (slice of maps) and it will panic if the input doesn't conform. Any other/better ideas on how to do this would be great

postprocess/postprocess.go
```diff
@@ -0,0 +1,161 @@
...
+
+	for i := 0; i < slice.Len(); i++ {
+		val := slice.Index(i).Elem()
+		if reflect.TypeOf(val).Kind() != reflect.Struct {
+			return Task{}, ErrBadPostprocessingPointsFormat
+		}
+
+		x := val.MapIndex(reflect.ValueOf("X"))
+		y := val.MapIndex(reflect.ValueOf("Y"))
+		task.Points = append(task.Points, r3.Vector{X: x.Elem().Float(), Y: y.Elem().Float()})
```
kim-mishra, line 58: Should we except Z input at all here?

postprocess/postprocess.go
```diff
@@ -0,0 +1,161 @@
...
+
+	// Initalize updatedData with new points
+	*updatedData = make([]byte, buf.Len())
+	updatedReader := bytes.NewReader(buf.Bytes())
+	updatedReader.Read(*updatedData)
+
+	return nil
+}
+
+func updatePointCloudWithRemovedPoints(updatedData *[]byte, points []r3.Vector) error {
```
kim-mishra, line 125: Not sure what the best way to do this is - the way it is currently implemented the user would have to input the **exact** point they want to be removed. Maybe we should instead take in a point and radius and remove any points within that circle?

postprocess/postprocess.go
```diff
@@ -0,0 +1,228 @@
...
+		if !ok {
+			return Task{}, ErrYNotProvided
+		}
+
+		yFloat, ok := y.(float64)
+		if !ok {
+			return Task{}, ErrXNotFloat64
+		}
+
+		task.Points = append(task.Points, r3.Vector{X: xFloat, Y: yFloat})
```
kim-mishra, line 104: Should we handle Z/make sure a user did not provide it or is this sufficient?

postprocess/postprocess.go
```diff
@@ -0,0 +1,228 @@
...
+
+const (
+	// Add is the instruction for adding points.
+	Add Instruction = iota
+	// Remove is the instruction for removing points.
+	Remove = iota
+)
+
+const (
+	fullConfidence = 100
```
kim-mishra, line 24: Right now i set all the added points to have 100% confidence. I did this because I figured if a user added it that number made sense. I could see an argument for having this be configurable but opted for this to simplify the `DoCommand`. What to others think?

postprocess/postprocess.go
```diff
@@ -0,0 +1,228 @@
...
+	// confirm iterate did not have to end early
+	if updatedPC.Size() != pc.Size()-len(points) {
+		/*
+			Note: this condition will occur if:
+				- points in task do not exist in the point cloud
+				- points have already been removed
+				- some error occurred while copying valid points
+		*/
+		return ErrRemovingPoints
+	}
```
kim-mishra, line 211: Is this worth keeping?

postprocess/postprocess.go
JohnN193: I think having a default sized radius would be fine for now.

If we wanted to cover more options then I’m wondering if there’s a good way to let users define a radius or a rectangle as options, depending on what gets sent in the do command

postprocess/postprocess.go
JohnN193: I think using full confidence makes sense. Cannot think of a use case where we would want it to be different

postprocess/postprocess.go
JohnN193: I think not handling z is sufficient

postprocess/postprocess.go
kim-mishra: updated to have a default radius of 1 cm for now

postprocess/postprocess.go
```diff
@@ -0,0 +1,230 @@
...
+	for _, point := range points {
+		/*
+			Viam expects pointcloud data with fields "x y z" or "x y z rgb", and for
+			this to be specified in the pointcloud header in the FIELDS entry. If color
+			data is included in the pointcloud, Viam's services assume that the color
+			value encodes a confidence score for that data point. Viam expects the
+			confidence score to be encoded in the blue parameter of the RGB value, on a
+			scale from 1-100.
+		*/
+		err := pc.Set(point, pointcloud.NewColoredData(color.NRGBA{B: fullConfidence}))
```
JohnN193, line 155: [q]Would we be interested in adding another color here to mark the pointcloud as an added obstacle? 

the prime component currently overwrites all colors to be grayscale based on the blue channel, so actually visualizing the obstacles would require a change in the prime component to handle the new channel but I think it would be a cool visual

JohnN193 (approved):
LGTM, theres a few questions floating around but I think everything here looks good!

postprocess/postprocess.go
randhid: Will it be more than a day's work? If so yep. If it's complicated, we'll do in a future date when we more formally do this. Save this idea, it's good.

postprocess/postprocess.go
```diff
@@ -0,0 +1,231 @@
...
+	ErrXNotFloat64 = errors.New("could not parse provided X as a float64")
+
+	// ErrYNotProvided denotes that a Y value was not provided.
+	ErrYNotProvided = errors.New("could X not provided")
+
+	// ErrYNotFloat64 denotes that an Y value is not a float64.
+	ErrYNotFloat64 = errors.New("could not parse provided X as a float64")
+
+	// ErrRemovingPoints denotes that something unexpected happened during removal.
+	ErrRemovingPoints = errors.New("unexpected number of points after removal")
```
randhid, line 60: Do any of these need to be exported?

viam_cartographer.go
```diff
@@ -34,6 +35,10 @@ var (
 	ErrClosed = errors.Errorf("resource (%s) is closed", Model.String())
 	// ErrUseCloudSlamEnabled denotes that the slam service method was called while use_cloud_slam was set to true.
 	ErrUseCloudSlamEnabled = errors.Errorf("resource (%s) unavailable, configured with use_cloud_slam set to true", Model.String())
+	// ErrNoPostprocessingToUndo denotes that the points have not been properly formatted.
+	ErrNoPostprocessingToUndo = errors.New("there are no postprocessing tasks to undo")
+	// ErrBadPostprocessingPointsFormat denotest that the postprocesing points have not been correctly provided.
+	ErrBadPostprocessingPointsFormat = errors.New("invalid postprocessing points format")
```
randhid, line 41: same with these and the constants in 53-57, if they can be unexported, that's best.

randhid (changes requested):

postprocess/postprocess.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+
+	buf := bytes.Buffer{}
+	err = pointcloud.ToPCD(updatedPC, &buf, pointcloud.PCDBinary)
+	if err != nil {
+		return err
+	}
+
+	// Overwrite updatedData with new points
+	*updatedData = make([]byte, buf.Len())
```
randhid, line 223: should we check if nil? What happens if it is?

postprocess/postprocess.go
kim-mishra: yes to be used in tests

postprocess/postprocess.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+
+	task := Task{}
+	for _, point := range pointSlice {
+		pointMap, ok := point.(map[string]interface{})
+		if !ok {
+			return Task{}, ErrPointNotAMap
+		}
+
+		x, ok := pointMap[xKey]
```
randhid, line 86: I think you could
```suggestion
		x, ok := pointMap[xKey].(float64)
```
To make it briefer

postprocess/postprocess.go
kim-mishra: if what is nil?

randhid: `*updatedData`, which we pass into Read. Come to think fo it can `buf.Len()` ever be zero?

viam_cartographer.go
kim-mishra: i can't use them in the viam cartographer tests if they are not exported

randhid: ah, since it is a `_test` package. Boooo

postprocess/postprocess.go
kim-mishra: I tested this in fmt.Print(make([]byte, 0) == nil) and it it is false

randhid: I think there's another one of these higher up, don't forget that one too!

postprocess/postprocess.go
```diff
@@ -0,0 +1,229 @@
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func updatePointCloudWithRemovedPoints(updatedData *[]byte, points []r3.Vector) error {
+	if updatedData == nil {
+		return errors.New("cannot provide nil udpated data")
```
randhid, line 175: This can also be a common error in this package for now.

randhid (approved):
One last cleanup nit, but LGTM! Love the processed map!

PR #312, opened 2023-12-29
add missing include

JohnN193:


sensorprocess/sensorprocess.go
```diff
@@ -34,7 +34,7 @@ func (config *Config) StartOfflineSensorProcess(ctx context.Context) bool {
 	lidarReading, err := config.Lidar.TimedLidarReading(ctx)
 	if err != nil {
 		config.Logger.Warn(err)
-		return errors.Is(err, replaypcd.ErrEndOfDataset)
+		return strings.Contains(err.Error(), replaypcd.ErrEndOfDataset.Error())
```
randhid, line 37: okay we should totally have the end of dataset logic be handled better, this is too brittle, I'm assuming it was sent over with an rpc wrapper and couldn't be compared using `errors.Is'? I wonder if we could have replay metadata be exposed somewhere.

randhid (approved):

PR #322, opened 2024-01-22
RSDK-6397 - create do Command to get positions from carto

kim-mishra:
### Adds 3 Commands to the `DoCommand`:
1. `PositionHistoryGetCommand = "position_history_get"`
      * Gets pointcloud representation of every position the robot has been in during a call to `Position`
2. `PositionHistoryEnableCommand = "position_history_enable"`
      * sets a boolean value that the FE checks and uses to determine whether or not to render the points, if this became a documented real feature we could just store this boolean in the FE and use a toggle button there to set it 
3. `PositionHistoryValueCommand = "position_history_value"`
      * Gets the boolean that determines whether or not the FE should display the position history
      
      
### Related PRS:
* RDK RC changes that call this DoCommand (calls command (3) & if it is true calls (1) to get the actual position and inputs to the prime component): https://github.com/viamrobotics/rdk/pull/3460
* Prime changes (combines path and map point cloud, path points have max value on the G and that are set to turquoise using that fact.: https://github.com/viamrobotics/prime/pull/469

https://github.com/viamrobotics/viam-cartographer/assets/121991867/1fa8dd4b-14d0-4abe-b680-095385089072

### DIFF
--- a/viam_cartographer.go
+++ b/viam_cartographer.go
@@ -5,6 +5,8 @@ package viamcartographer
 import (
 	"bytes"
 	"context"
+	"image/color"
+	"math"
 	"os"
 	"path/filepath"
 	"strconv"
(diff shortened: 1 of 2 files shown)


randhid (changes requested):
I would like the front end changes to show smaller points for the path (let me know if that is part of the RC card PR) so we don't color all over the map with blue. Also, this could get messy and large really quickly without a way to clear the history as well. Do you see any ways for us to deal with that in the back or front end?

PR #325, opened 2024-02-05
RSDK-4863 & RSDK-4670 - Integration tests for odometry & logging improvements

kkufieta:
Tickets:
* https://viam.atlassian.net/browse/RSDK-4863
* https://viam.atlassian.net/browse/RSDK-4670

Done:
* Added new data for integration tests using odometry data
* Extended tests to test lidar and odometry data
* Improved logging

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -48,7 +48,7 @@ func TestIntegrationCartographer(t *testing.T) {
 	}{
 		// Online run with lidar only
 		{
-			description:     "online sensor mapping mode 2D",
+			description:     "online mapping mode 2D with lidar only",
 			online:          true,
 			imuEnabled:      false,
 			odometerEnabled: false,
@@ -56,7 +56,7 @@ func TestIntegrationCartographer(t *testing.T) {
 			subAlgo:         viamcartographer.Dim2d,
 		},
 		{
-			description:     "online sensor localizing mode 2D",
+			description:     "online localizing mode 2D with lidar only",
 			online:          true,
 			imuEnabled:      false,
 			odometerEnabled: false,
@@ -64,7 +64,7 @@ func TestIntegrationCartographer(t *testing.T) {
 			subAlgo:         viamcartographer.Dim2d,
 		},
 		{
-			description:     "online sensor updating mode 2D",
+			description:     "online updating mode 2D with lidar only",
 			online:          true,
 			imuEnabled:      false,
 			odometerEnabled: false,
@@ -73,7 +73,7 @@ func TestIntegrationCartographer(t *testing.T) {
 		},
 		// Offline run with lidar only
 		{
-			description:     "offline sensor mapping mode 2D",
+			description:     "offline mapping mode 2D with lidar only",
 			online:          false,
 			imuEnabled:      false,
 			odometerEnabled: false,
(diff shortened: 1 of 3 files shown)


integration_test.go
```diff
@@ -39,70 +39,104 @@ func TestIntegrationCartographer(t *testing.T) {
 	logger := logging.NewTestLogger(t)
 
 	cases := []struct {
-		description string
-		online      bool
-		imuEnabled  bool
-		mode        cartofacade.SlamMode
-		subAlgo     viamcartographer.SubAlgo
+		description     string
```
jeremyrhyde, line 42: @randhid We currently only do offline testing for lidar only as the online case for imu + lidar is more of a stress test than the offline case would be. If you would like every possible combination we can support this by changing the case structure to for loops that cover all possible combination of online [true, false], imuEnabled [tru, false]. odomEnabled [true, false].

integration_test.go
```diff
@@ -39,70 +39,104 @@ func TestIntegrationCartographer(t *testing.T) {
...
+			mode:            cartofacade.LocalizingMode,
+			subAlgo:         viamcartographer.Dim2d,
+		},
+		{
+			description:     "online updating mode 2D with lidar + odometer",
+			online:          true,
+			imuEnabled:      false,
+			odometerEnabled: true,
+			mode:            cartofacade.UpdatingMode,
+			subAlgo:         viamcartographer.Dim2d,
```
jeremyrhyde, line 139: I think it would be beneficial to have a case where both odom and imu data are being used

jeremyrhyde (commented):
@kkufieta can you attach a copy of an entry from the new movementsensor dataset? I wanted to see how the imu + odom data is being stored

integration_test.go
kkufieta: It was decided that it would be out-of-scope for the ticket.

integration_test.go
kkufieta: We'd have to refactor the integration tests for this, though, because offline only works for the lidar, not for the lidar + imu or odom or both.

kkufieta:
> @kkufieta can you attach a copy of an entry from the new movementsensor dataset? I wanted to see how the imu + odom data is being stored

[data.json](https://github.com/viamrobotics/viam-cartographer/files/14171626/data.json)

integration_test.go
jeremyrhyde: Refactoring would be simple, it would just involve skipping the blocking loop in the integration lidar/imu

kkufieta: Ah, you're right!

testhelper/integrationtesthelper.go
```diff
@@ -533,23 +662,42 @@ func mockLidarReadingsValid() error {
...
+	var data movementSensorData
+
+	if err = json.Unmarshal(byteValue, &data); err != nil {
+		t.Error("TEST FAILED TimedMovementSensorReading Mock failed to unmarshal json")
+		return movementSensorData{}, err
+	}
 
-	for mockDatasetScanner.Scan() {
-		mockDataset = append(mockDataset, mockDatasetScanner.Text())
+	if len(data.AngVelData) != len(data.LinAccData) &&
```
jeremyrhyde, line 691: [nit] 
```
expectedDataNum := len(data.AngVelData)
if expectedDataNum!= len(data.LinAccData) &&
		expectedDataNum != len(data.OrientationData) &&
		expectedDataNum != len(data.PosData) {
....

if expectedDataNum < NumMovementSensorData {
```

Also we may want to check that the `expectedDataNum >NumMovementSensorData` before comparing to all other data fields

jeremyrhyde (approved):
LGTM, one small nit

randhid (approved):

PR #329, opened 2024-03-12
[RSDK-6873] add edited map check to cartographer

JohnN193:
https://viam.atlassian.net/browse/RSDK-6873


viam_cartographer.go
```diff
@@ -62,6 +62,7 @@ const (
 	SuccessMessage = "success"
 	// PostprocessToggleResponseKey is the key sent back for the toggle postprocess command.
 	PostprocessToggleResponseKey = "postprocessed"
+	editedMapName                = "/edited-map.pcd"
```
randhid, line 65: Add a ticket to the Epic to call out that we're expecting the edited map packages to have this name from front end.

randhid (approved):
Love it!

viam_cartographer.go
```diff
@@ -270,6 +271,20 @@ func New(
...
+
+		packageDir = filepath.Clean(packageDir + editedMapName)
+		bytes, err := os.ReadFile(packageDir)
+		if err != nil && !errors.Is(err, os.ErrNotExist) {
+			return nil, err
+		}
+		if len(bytes) > 0 {
+			cartoSvc.editedMap = &bytes
+		}
+	}
```
JohnN193, line 286: so theres not really a good way to do this now, but at some point we should probably add a function to do this in rdk/slam supported sdks. Other option is documenting how slam-map packages workso they know how to do this.
cc @randhid

JohnN193: tbh i wonder if there was a way to have the sdk/server codes pull from the packages directly rather than interacting with the slam interface, but thats def a question for another day

randhid: We can expose helpers client side on the different sdks. However you're right it's heavily dependent on the implementer to implement. Which is fine I think.

randhid: I'd keep it carto-side fro now becuase we have no idea how this API works or behaves with visual slam and whether we need a non-pcd expansion for those maps.

.... or bump sensor slam.

testhelper/testhelper.go
```diff
@@ -247,7 +247,7 @@ func InitTestCL(t *testing.T, logger logging.Logger) func() {
 // InitInternalState creates the internal state directory witghin a temp directory
```
kkufieta, line 247: [opt]
```suggestion
// InitInternalState creates the internal state directory within a temp directory
```

kkufieta (approved):

PR #330, opened 2024-03-16
Try to create appimages with different model

kkufieta:


kkufieta:
Just needed to create the AppImages.

PR #335, opened 2024-04-10
[RSDK-7269] allow for empty config params and fix properties

JohnN193:
lets the strings not fail if empty strings are passed in for min and max range, and fixes properties while cloudslam is enabled

https://viam.atlassian.net/browse/RSDK-7269

### DIFF
--- a/viam_cartographer.go
+++ b/viam_cartographer.go
@@ -337,16 +337,24 @@ func parseCartoAlgoConfig(configParams map[string]string, logger logging.Logger)
 			}
 			cartoAlgoCfg.MissingDataRayLength = float32(fVal)
 		case "max_range":
+
 			fVal, err := strconv.ParseFloat(val, 32)
 			if err != nil {
 				return cartoAlgoCfg, err
 			}
 			cartoAlgoCfg.MaxRange = float32(fVal)
 		case "max_range_meters":
-			fVal, err := strconv.ParseFloat(val, 32)
-			if err != nil {
-				return cartoAlgoCfg, err
+			var fVal float64
+			var err error
+			if val == "" {
+				fVal = float64(defaultCartoAlgoCfg.MaxRange)
+			} else {
+				fVal, err = strconv.ParseFloat(val, 32)
+				if err != nil {
+					return cartoAlgoCfg, err
+				}
 			}
+
 			cartoAlgoCfg.MaxRange = float32(fVal)
 		case "min_range":
 			fVal, err := strconv.ParseFloat(val, 32)
@@ -355,9 +363,15 @@ func parseCartoAlgoConfig(configParams map[string]string, logger logging.Logger)
 			}
 			cartoAlgoCfg.MinRange = float32(fVal)
 		case "min_range_meters":
-			fVal, err := strconv.ParseFloat(val, 32)
-			if err != nil {
-				return cartoAlgoCfg, err
+			var fVal float64
+			var err error
+			if val == "" {
+				fVal = float64(defaultCartoAlgoCfg.MinRange)
+			} else {
+				fVal, err = strconv.ParseFloat(val, 32)
+				if err != nil {
+					return cartoAlgoCfg, err
+				}
 			}
 			cartoAlgoCfg.MinRange = float32(fVal)
 		case "max_submaps_to_keep":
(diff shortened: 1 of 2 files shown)


viam_cartographer.go
```diff
@@ -289,9 +289,13 @@ func New(
...
-			Named:        c.ResourceName().AsNamed(),
-			useCloudSlam: true,
-			logger:       logger,
+			Named:          c.ResourceName().AsNamed(),
+			useCloudSlam:   true,
+			logger:         logger,
+			lidar:          timedLidar,
+			movementSensor: timedMovementSensor,
+			enableMapping:  optionalConfigParams.EnableMapping,
+			existingMap:    optionalConfigParams.ExistingMap,
```
JohnN193, line 298: basically we were not passing anything into the cartographer service struct when cloudslam was enabled, so GetProperties was failing to actually get the properties of the slam service

viam_cartographer.go
```diff
@@ -337,16 +341,24 @@ func parseCartoAlgoConfig(configParams map[string]string, logger logging.Logger)
 			}
 			cartoAlgoCfg.MissingDataRayLength = float32(fVal)
 		case "max_range":
+
```
penguinland, line 344: Mild preference to remove this line again, and never start a new indentation level with a blank line.

viam_cartographer.go
```diff
@@ -337,16 +341,24 @@ func parseCartoAlgoConfig(configParams map[string]string, logger logging.Logger)
...
 			fVal, err := strconv.ParseFloat(val, 32)
 			if err != nil {
 				return cartoAlgoCfg, err
 			}
 			cartoAlgoCfg.MaxRange = float32(fVal)
 		case "max_range_meters":
```
JohnN193, line 350: only making this change for `max_range_meters` and `min_range_meters` because they are the only parameters that automatically get added to the config on the r2d2 slam card with the chance of being empty

viam_cartographer.go
```diff
@@ -337,16 +341,24 @@ func parseCartoAlgoConfig(configParams map[string]string, logger logging.Logger)
...
-			fVal, err := strconv.ParseFloat(val, 32)
-			if err != nil {
-				return cartoAlgoCfg, err
+			var fVal float64
+			var err error
+			if val == "" {
+				fVal = float64(defaultCartoAlgoCfg.MaxRange)
```
penguinland, line 354: :+1: Very good catch here!

viam_cartographer.go
```diff
@@ -355,9 +367,15 @@ func parseCartoAlgoConfig(configParams map[string]string, logger logging.Logger)
...
-			if err != nil {
-				return cartoAlgoCfg, err
+			var fVal float64
+			var err error
+			if val == "" {
+				fVal = float64(defaultCartoAlgoCfg.MinRange)
+			} else {
+				fVal, err = strconv.ParseFloat(val, 32)
+				if err != nil {
+					return cartoAlgoCfg, err
```
penguinland, line 377: You've copied and pasted the same fix twice. Would it be worth making this a helper function? That would make it clearer that it's the same code in both places, and if you have future bugs to fix or future features to add, they only need to be changed in one location (no chance of forgetting about the second one). 

but I don't feel strongly: leave as-is if you think that's the right choice.

penguinland (approved):
LGTM! All my feedback is optional.

viam_cartographer.go
```diff
@@ -355,9 +367,15 @@ func parseCartoAlgoConfig(configParams map[string]string, logger logging.Logger)
...
 		case "min_range_meters":
-			fVal, err := strconv.ParseFloat(val, 32)
-			if err != nil {
-				return cartoAlgoCfg, err
+			var fVal float64
+			var err error
+			if val == "" {
+				fVal = float64(defaultCartoAlgoCfg.MinRange)
+			} else {
+				fVal, err = strconv.ParseFloat(val, 32)
```
penguinland, line 375: ...line 373 has a float64, but line 375 seems to expect a 32-bit float. Are you sure about this? ("Yes" is a very believable answer!)

viam_cartographer.go
penguinland: Something like:
```go
func parseOrDefault(val string, default float64) (float64, error) {
    if val == "" {
        return default, nil
    }
    return strconv.ParseFloat(val, 32)
}
```
and then you can pass in the MaxRange or the MinRange as the default in the two places it would be used. Admittedly, it's not as long a helper function as I expected, so maybe not worth the hassle.

penguinland (approved):

viam_cartographer.go
JohnN193: making it a helper function lets me put it in every other relevant case, so def a good callout

viam_cartographer.go
JohnN193: yeah, the issue is ParseFloat always returns a float64, evene when you tell it to use 32 bits. Setting up the help functions to handle this better tho

PR #342, opened 2024-10-09
[WORKFLOW] Updating go dependencies

JohnN193:
tests don't seem to be running on these so trying to figure out why

changes being made

bump api/rdk (and go to 23)
update the linter config for the repo(to help [match rdk](https://github.com/viamrobotics/rdk/blob/main/etc/.golangci.yaml))
update bump dependency workflow
remove some dead code related to generating c++ grpc code


.github/workflows/bump_dependencies.yml
```diff
@@ -43,6 +44,7 @@ jobs:
           title: Automated Go Dependencies Update
           body: This is an auto-generated PR to update go dependencies. Please confirm tests are passing before merging.
           reviewers: JohnN193
+          draft: always-true
```
martha-johnston, line 47: interesting, what does this do?

martha-johnston (approved):

.github/workflows/bump_dependencies.yml
JohnN193: whenever this workflow makes a pr or updates a pr, it should set that pr to be in draft mode. hoping to use this to be able to trigger tests to run in the pr without having to create a different pr, though i might need to make another change to other workflows first before it actually works

PR #350, opened 2024-12-20
RSDK-9547

martha-johnston:
fix artifact download


martha-johnston:
no idea how to test locally if the deploy workflow will work

martha-johnston:
based on these docs, seems like there's multiple options? unsure which is the best choice https://github.com/actions/download-artifact/blob/main/docs/MIGRATION.md

JohnN193:
oh you need to have the pr branch be a branch off of main, not on your fork

PR #351, opened 2024-12-20
RSDK-9547: fix artifact v4 upload

martha-johnston:


martha-johnston:
I think it's going to fail no matter what because it's using the workflows in main which are broken

